PR #4, opened 2019-03-22
DROOLS-3786: Polyglot

evacchi:
- Add an example with JS interop on Graal
- Comes with a blog post (to be published)
- Please squash before merge


pom.xml
```diff
@@ -46,6 +46,7 @@
   </repositories>
   <modules>
     <module>drools-quarkus</module>
+    <module>drools-polyglot</module>
```
mswiderski: @evacchi could you add suffix -example to the modules so they are bit more meaningful ... on maven repo side. If someone sees the "drools-polyglot" might not know it's an example

evacchi: done

PR #6, opened 2019-04-03
JBPM-8397 - Update swagger maven plugin for submarine generated apps

tsurdilo:


mswiderski (approved):

evacchi:
fyi there's a swagger-ui quarkus exension

mswiderski:
we currently don't include swagger ui at all only swagger definition but that is good to know, thanks

PR #7, opened 2019-04-10
JBPM-8320: Redesign of ProcessRuntime API

evacchi:
fix examples: disable dependencyInjection for SpringBoot and disable generateProcessModel for Drools


mswiderski:
closing this one as it's no longer needed, spring boot sample is already adjusted and the generate process model is not needed as the process model is generated only if at least on BPMN file is found

PR #8, opened 2019-04-11
Initial implementation of OpenShift tests for Submarine examples

sutaakar:
@mswiderski Couple of questions:
- Should I add licence headers into the files?
- Right now tests are executed with basic build (mvn clean install). Considering that they need external resource (OpenShift) and takes quite a lot of time, would it have sense to move them to some profile?

### DIFF
--- a/openshift-example-tests/.gitignore
+++ b/openshift-example-tests/.gitignore
@@ -32,6 +32,3 @@ jdbc_driver.jar
 db-settings.xml
 
 quarkus.log
-
-# XTF properties files
-test.properties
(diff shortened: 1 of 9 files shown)


mswiderski:
> Should I add licence headers into the files?

yes

> Right now tests are executed with basic build (mvn clean install). Considering that they need external resource (OpenShift) and takes quite a lot of time, would it have sense to move them to some profile?

yes, let's have a dedicated profile for them

sutaakar:
Licence headers added.
Integration tests skipped by default. They can be enabled by running Maven command with -DskipITs=false. Thinking that it may be better solution than using artificial profile just to enable tests.
@mswiderski What do you think about that?

mswiderski:
@sutaakar yes, I'd add a profile for it that would make it more explicit

sutaakar:
Added profile "openshift" enabling the tests.

mswiderski (approved):
@sutaakar is this ready to be merged?

sutaakar:
Yes.

PR #10, opened 2019-04-11
fixed 'org.jboss.resteasy.core.NoMessageBodyWriterFoundFailure: Could…

hwurht:
… not find MessageBodyWriter for response object of type' when running persons/all & persons/adult endpoints


mswiderski (approved):
@mariofusco @evacchi any objections?

evacchi:
I think that should be fine

PR #36, opened 2019-06-12
DROOLS-4160: Enable DRL codegen by default, rename config properties

evacchi:
depends on https://github.com/kiegroup/kogito-runtimes/pull/68


mswiderski (approved):

mswiderski:
looks like rebase is needed?

PR #41, opened 2019-07-02
[DROOLS-4241] upgrade rule unit example to new API

mariofusco:


evacchi:
it should be fine, but CI is red..

PR #42, opened 2019-07-03
KOGITO-20 - Unit of work support for process runtime to control execu…

mswiderski:
…tion

depends https://github.com/kiegroup/kogito-runtimes/pull/82

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,6 +21,7 @@
   <properties>
     <surefire.forkCount>1</surefire.forkCount>
     <alphanetworkCompilerEnabled>false</alphanetworkCompilerEnabled>
+    <version.org.junit>5.4.2</version.org.junit>
   </properties>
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -21,6 +21,7 @@
   <properties>
     <surefire.forkCount>1</surefire.forkCount>
     <alphanetworkCompilerEnabled>false</alphanetworkCompilerEnabled>
+    <version.org.junit>5.4.2</version.org.junit>
```
mswiderski, line 24: @triceo I had to override version of junit as 5.5... is just silently skipping tests in this module .. at least those based on quarkus. Are you aware of that?

triceo: I am not. I will look into it. Thanks for pointing this out!

evacchi (approved):

PR #47, opened 2019-07-17
[KOGITO-56] first implementation of automatic rest endpoint generatio…

mariofusco:
…n for queries


drools-quarkus-unit-example/src/main/java/org/kie/kogito/queries/AdultUnit.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public int getAdultAge() {
+        return adultAge;
+    }
+
+    public void setAdultAge( int adultAge ) {
+        this.adultAge = adultAge;
+    }
+
+    public void setPersonList( List<Person> personList ) {
```
evacchi: it would be great if we can do this with "magic" mapping via jsonb

drools-quarkus-unit-example/src/main/resources/META-INF/kmodule.xml
```diff
@@ -1,12 +1,9 @@
...
-<!--    <kbase name="simpleKB" packages="org.drools.simple.project">-->
-<!--        <ksession name="simpleKS" default="true"/>-->
-<!--    </kbase>-->
-<!--    <kbase name="canDrinkKB" packages="org.drools.simple.candrink">-->
-<!--        <ksession name="canDrinkKS"/>-->
-<!--    </kbase>-->
     <kbase name="alertingKB" packages="org.kie.kogito.rules.alerting">
         <ksession name="alertingKS"/>
     </kbase>
+    <kbase name="queriesKB" packages="org.kie.kogito.queries">
```
evacchi, line 6: memo: we may be already able to do this without kmodule

evacchi (approved):

PR #56, opened 2019-07-26
Use admin user to install Kogito Operator for tests

sutaakar:
The ordinary user doesn't have rights for this.
@ricardozanini Could you please review?

### DIFF
--- a/openshift-example-tests/src/main/java/org/kogito/examples/openshift/deployment/OperatorDeployer.java
+++ b/openshift-example-tests/src/main/java/org/kogito/examples/openshift/deployment/OperatorDeployer.java
@@ -18,11 +18,11 @@ package org.kogito.examples.openshift.deployment;
 import java.net.MalformedURLException;
 import java.net.URL;
 
-import cz.xtf.core.openshift.OpenShiftBinary;
-import cz.xtf.core.openshift.OpenShifts;
 import io.fabric8.kubernetes.api.model.ServiceAccount;
 import io.fabric8.kubernetes.api.model.apiextensions.CustomResourceDefinition;
 import io.fabric8.kubernetes.api.model.apps.Deployment;
+import io.fabric8.kubernetes.api.model.rbac.Role;
+import io.fabric8.kubernetes.api.model.rbac.RoleBinding;
 import io.fabric8.kubernetes.client.dsl.NonNamespaceOperation;
 import io.fabric8.kubernetes.client.dsl.Resource;
 import org.kogito.examples.openshift.Project;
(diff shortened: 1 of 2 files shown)


openshift-example-tests/src/main/java/org/kogito/examples/openshift/deployment/OperatorDeployer.java
```diff
@@ -53,29 +54,40 @@
...
+            project.getAdmin().customResourceDefinitions().create(customResourceDefinition);
+        }
+    }
+
     private static void createServiceAccountInProject(Project project) {
         logger.info("Creating service account in project '" + project.getName() + "' from " + OPERATOR_SERVICE_ACCOUNT);
-        ServiceAccount serviceAccount = project.getMaster().serviceAccounts().load(getResource(OPERATOR_SERVICE_ACCOUNT)).get();
-        project.getMaster().createServiceAccount(serviceAccount);
+        ServiceAccount serviceAccount = project.getAdmin().serviceAccounts().load(getResource(OPERATOR_SERVICE_ACCOUNT)).get();
+        project.getAdmin().createServiceAccount(serviceAccount);
```
ricardozanini, line 78: The Operator Service Account will be created within the Operator https://issues.jboss.org/browse/KOGITO-66. I understand that you need this now, but you should remember to remove it as soon as I fix this JIRA. :p

openshift-example-tests/src/main/java/org/kogito/examples/openshift/deployment/OperatorDeployer.java
```diff
@@ -53,29 +54,40 @@
...
     }
 
-    private static void createRoleInProject(Project project, OpenShiftBinary masterBinary) {
+    private static void createRoleInProject(Project project) {
         logger.info("Creating role in project '" + project.getName() + "' from " + OPERATOR_ROLE);
-        masterBinary.execute("create", "-f", OPERATOR_ROLE);
+        Role role = project.getAdmin().rbac().roles().load(getResource(OPERATOR_ROLE)).get();
+        project.getAdmin().rbac().roles().inNamespace(project.getName()).create(role);
```
ricardozanini, line 84: See above, same for the bindings. We could end up with duplicated content. Or this user need to be created to run the Operator?

ricardozanini (changes requested):

openshift-example-tests/src/main/java/org/kogito/examples/openshift/deployment/OperatorDeployer.java
sutaakar: These steps are actually required to install Kogito Operator itself. They are not related to actions done by Kogito Operator when building an application using CRs.
This class is responsible for installation of Kogito Operator. Current approach is to install it "manually" by creating all needed object by ourself - as Kogito operator is not yet published in catalog.

ricardozanini: Sure!

ricardozanini (approved):

mswiderski (approved):

PR #58, opened 2019-07-31
KOGITO-60 - Runtime persistence based on Infinispan, KOGITO-16 - Change process instance/node instance/work item id from Long to String to not depend on data store for id generation

mswiderski:
depends on
https://github.com/kiegroup/kogito-bom/pull/43
https://github.com/kiegroup/kogito-runtimes/pull/104


jbpm-quarkus-example/README.md
```diff
@@ -41,6 +41,23 @@ To run the generated native executable, generated in `target/`, execute
 ./target/jbpm-quarkus-example-{version}-runner
 ```
 
+### Running with persistence enabled
+
+Kogito support runtime persistence that is backed by Inifinispan. So to be able to enable this you need to have
```
cristianonicolai: Should we add a basic instruction on how to  get the Infinistan server and run? like download link, unzip, .startServer ?

mswiderski: since at the moment we don't have any special requirements it should be enough to rely on Infinispan docs itself, isn't it?

cristianonicolai: sure, as simple as a link to the website could be enough then.

mswiderski: updated with short info where to download it and which version

PR #60, opened 2019-08-05
[KOGITO-115] - reading properties during process execution

ricardozanini:
This PR depends on: https://github.com/kiegroup/kogito-cloud/pull/28

### DIFF
--- a/onboarding-example/onboarding/src/main/java/org/kie/kogito/examples/onboarding/DecisionTaskWorkItemHandler.java
+++ b/onboarding-example/onboarding/src/main/java/org/kie/kogito/examples/onboarding/DecisionTaskWorkItemHandler.java
@@ -10,27 +10,23 @@ import org.kie.kogito.cloud.workitems.ServiceInfo;
 
 public class DecisionTaskWorkItemHandler extends DiscoveredServiceWorkItemHandler {
 
-    private String namespace = System.getenv("NAMESPACE");
-
     public DecisionTaskWorkItemHandler() {
         if ("true".equalsIgnoreCase(System.getProperty("local"))) {
-            this.serviceEndpoints.put("id", new ServiceInfo("http://localhost:8081/id", null));
-            this.serviceEndpoints.put("department", new ServiceInfo("http://localhost:8081/department", null));
-            this.serviceEndpoints.put("employeeValidation", new ServiceInfo("http://localhost:8081/employeeValidation", null));
-            this.serviceEndpoints.put("vacationDays", new ServiceInfo("http://localhost:8082/vacationDays", null));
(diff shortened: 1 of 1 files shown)


onboarding-example/onboarding/src/main/java/org/kie/kogito/examples/onboarding/DecisionTaskWorkItemHandler.java
```diff
@@ -9,29 +9,33 @@
 import org.kie.kogito.cloud.workitems.ServiceInfo;
 
 public class DecisionTaskWorkItemHandler extends DiscoveredServiceWorkItemHandler {
-
-    private String namespace = System.getenv("NAMESPACE");
+    
+    private boolean localServicesAdded;
```
mswiderski: why do we need this flag?

onboarding-example/onboarding/src/main/java/org/kie/kogito/examples/onboarding/DecisionTaskWorkItemHandler.java
```diff
@@ -9,29 +9,33 @@
...
+            this.addServices("paymentDate", new ServiceInfo("http://localhost:8082/paymentDate", null));
+            this.localServicesAdded = true;
+        } 
     }
 
     @Override
     public void executeWorkItem(WorkItem workItem, WorkItemManager manager) {
-
-        Map<String, Object> results = discoverAndCall(workItem, namespace, "Decision", HttpMethods.POST);
+        this.addLocalServices();
```
mswiderski: why do we need to do it at execute work item and not in the constructor? this is just static configuration so should not cause any additional look ups...

ricardozanini: I don't like this approach either. The problem is that the map can't hold the items added to it in the constructor, that's why we have to do this check every time. It's a different problem than the `System.getenv` one.

onboarding-example/onboarding/src/main/java/org/kie/kogito/examples/onboarding/DecisionTaskWorkItemHandler.java
ricardozanini: To not add the items in every process execution

onboarding-example/onboarding/src/main/java/org/kie/kogito/examples/onboarding/DecisionTaskWorkItemHandler.java
mswiderski: what's the reason for this?

mswiderski (approved):

PR #63, opened 2019-08-07
jbpm polyglot example

tsurdilo:
simple jBPM polyglot example. it includes a process with a rule task which calls applicant rules based on the entered age.

graalvm is needed to run this example.

see the readme and the images for info on how to start the nodejs server and enter in simple applicant data to be evaluated.


jbpm-polyglot-example/pom.xml
```diff
@@ -0,0 +1,159 @@
...
+            <executions>
+              <execution>
+                <phase>package</phase>
+                <goals>
+                  <goal>exec</goal>
+                </goals>
+              </execution>
+            </executions>
+            <configuration>
+              <executable>/Library/Java/JavaVirtualMachines/graalvm-ce-19.1.1/Contents/Home/bin/node</executable>
```
mswiderski: I guess we need this to be parametrised or just assumed it's on path

jbpm-polyglot-example/src/main/node/app.js
```diff
@@ -0,0 +1,67 @@
+const Applicant = Java.type('org.kie.kogito.examples.polyglot.Applicant');
+const Application = Java.type('org.kie.kogito.examples.Application');
+const ProcessComponent = Java.type('org.kie.kogito.examples.polyglot.ProcessComponent');
```
mswiderski: any reason why not to use the generated Process type here? Then we could most likely avoid the ProcessComponent completely...

mswiderski (commented):
@evacchi you might want to look at it as well

evacchi (commented):
interesting; however, from the title of the example, I was expecting this to use the jBPM DSL instead of the XML definition: cf https://github.com/evacchi/kie-graal-boilerplate/blob/master/jbpm.js

food for thought: if we embed the Python interpreter (supported: Graal is truly polyglot) we could even come up with a Kube/YAML-friendly format.

tsurdilo:
@evacchi thanks for the comment. the way i see this as user of kogito is yeah it's cool to be able to write dsl for bpmn2 or drools, but why? To me kogito is about easy integration, and in case of this examples using graalvm polyglot features to simply integrate kogito with node, python (will add example for it too) etc. I don't think I would want to write my business processes in javascript even if I could I think :) One nice feature we maybe should promote is ok we can integrate kogito with any UI technology without graalvm, but one feature I found nice with this is that since its executed in vm there has to be no callbacks, or extra rest calls, webhooks, reactive libs or anything to get the results of rules, processes back to the user. That and the fact that we can simply use the generated code in any language the user may prefer to code in. Hope I make sense :)

jbpm-polyglot-example/pom.xml
tsurdilo: sorry forgot to remove, changed it to just "node" as in the drools example and added comment on readme to put in full path if needed.

jbpm-polyglot-example/src/main/node/app.js
tsurdilo: was just for convenience (less js code). updated , removed ProcessComponent

jbpm-polyglot-example/src/main/node/app.js
```diff
@@ -0,0 +1,73 @@
...
+
+function getApplicantProcessInstance(applicant) {
+
+    var kogitoApp = new ApplicationType();
+    var applicantProcess = kogitoApp.processes().processById("applicantprocess");
+    var dataModel = applicantProcess.createModel();
+
+    var dataMap = new HashMapType();
+    dataMap.put("applicant", applicant);
+    dataModel.fromMap(dataMap);
```
mswiderski: what I had in mind is to use already generated process class and data model so no need to use generic api

```
const ApplicationModel= Java.type('org.kogito.examples.polyglot.ApplicantprocessModel');

var model = new ApplicationModel();
model.setApplicant(applicant);

var process = kogitoApp.processes().createApplicantprocessProcess() // this is based on process id in bpmn2 so might be worth to make it bit more meaningful

process.createInstance(model);

```

tsurdilo: ok sorry didnt get it at first. updated.

jbpm-polyglot-example/src/main/node/app.js
```diff
@@ -0,0 +1,73 @@
...
+
+function getApplicantProcessInstance(applicant) {
+
+    var kogitoApp = new ApplicationType();
+
+    var model = new ApplicantProcessModelType();
+    model.setApplicant(applicant);
+
+    var process = kogitoApp.processes().createApplicantprocessProcess();
+    return process.createInstance(model);
```
mswiderski, line 71: now this is nice and simple, thanks

mswiderski (approved):
looks good to me

optionally we could change process id so the api does not look like ApplicantprocessProcess maybe changing process id to applicants will make it ApplicantProcess and then ApplicantModel which looks better imo. Maybe add license headers to the java classes.

tsurdilo:
added license headers

PR #65, opened 2019-08-08
Force kubernetes client to use last version

radtriste:
beacuse of a bug on project creation in Openshift 4:
https://github.com/fabric8io/kubernetes-client/issues/1631

set tmp/ folder in .gitignore as it can be created by cz-xtf library if
no oc binary path given and is not deleted ...

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,6 +21,8 @@
   <properties>
     <surefire.forkCount>1</surefire.forkCount>
     <alphanetworkCompilerEnabled>false</alphanetworkCompilerEnabled>
+    <!-- Force kubernetes client version as long as cz-xtf library did not integrate it, due to bug https://github.com/fabric8io/kubernetes-client/issues/1631 -->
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -21,6 +21,8 @@
   <properties>
     <surefire.forkCount>1</surefire.forkCount>
     <alphanetworkCompilerEnabled>false</alphanetworkCompilerEnabled>
+    <!-- Force kubernetes client version as long as cz-xtf library did not integrate it, due to bug https://github.com/fabric8io/kubernetes-client/issues/1631 -->
+    <version.kubernetes-client>4.4.0</version.kubernetes-client>
```
sutaakar, line 25: It would be better to put the version to Kogito BOM, so it is propagated to all projects.

radtriste:
Created https://github.com/kiegroup/kogito-bom/pull/51 to update kogito-bom

PR #66, opened 2019-08-09
fixed resource package for generating swagger doc

mswiderski:


sutaakar:
Swagger is correctly generated now, thanks.

sutaakar (approved):

PR #67, opened 2019-08-09
small fix for jbpm polyglot example

tsurdilo:


tsurdilo:
@mswiderski i removed the rules part because of today there was some change with rule integration that i cannot yet figure out. just changing process to check age instead of calling rule task.

PR #72, opened 2019-08-20
KOGITO-102: jBPM Rule Unit integration

evacchi:
Brainstorming integration API, TBD runtime support.

- `PersonRules` is a rule unit (silly rule for now, we will make it better)
  - `persons` is a `DataStore<Person>`
- PersonsProcess has a `Person person` variable
    - it calls into rule unit called `PersonRules` using a BusinessRuleTask 
    - input: maps `person` to data source `persons` (i.e. it **inserts** it implicitly inside)
    - output: maps `person` to data source `persons` (i.e. it takes one value -- this should be probably discussed: probably too much magic) -- notice we allow this because data sources may be seen as queries

### DIFF
--- a/jbpm-quarkus-example/src/main/java/org/kie/kogito/examples/PersonValidationService.java
+++ b/jbpm-quarkus-example/src/main/java/org/kie/kogito/examples/PersonValidationService.java
@@ -1,0 +1,14 @@
+package org.kie.kogito.examples;
+
+import org.kie.kogito.examples.demo.Person;
+import org.kie.kogito.rules.DataSource;
+import org.kie.kogito.rules.DataStore;
+import org.kie.kogito.rules.RuleUnitMemory;
+
+public class PersonValidationService implements RuleUnitMemory {
+    private DataStore<Person> persons = DataSource.createStore();
+
+    public DataStore<Person> getPersons() {
+        return persons;
+    }
+}
--- a/jbpm-quarkus-example/src/main/resources/org/kie/kogito/examples/PersonValidationService.drl
+++ b/jbpm-quarkus-example/src/main/resources/org/kie/kogito/examples/PersonValidationService.drl
@@ -1,12 +1,11 @@
-package org.acme.kogito
+package org.kie.kogito.examples
+unit PersonValidationService
 
 import org.kie.kogito.examples.demo.Person;
 
-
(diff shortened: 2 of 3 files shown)


jbpm-quarkus-example/src/main/resources/org/kie/kogito/examples/person-rules.drl
```diff
@@ -1,14 +0,0 @@
-package org.acme.kogito
-
-import org.kie.kogito.examples.demo.Person;
-
-
-rule "Is adult" ruleflow-group "person"
```
mswiderski, line 6: why removing this? it's for another process not related to orders :)

jbpm-quarkus-example/src/main/resources/org/kie/kogito/examples/orderItems.bpmn
```diff
@@ -59,7 +59,13 @@
...
       <bpmn2:incoming>SequenceFlow_4</bpmn2:incoming>
     </bpmn2:endEvent>
-    <bpmn2:serviceTask id="ServiceTask_1" name="Calculate total" implementation="java" operationRef="Operation_1">
+    <!-- 
+      temporarily, for interim tooling support we use 'unit:' prefix in 'drools:ruleFlowGroup'
+      name of the unit may be prefixed by package 
+      'drools:ruleFlowGroup="unit:CalculationService"' or 'drools:ruleFlowGroup="unit:org.kie.kogito.examples.CalculationService"'';
+      in the future it might become 'drools:ruleUnit="CalculationService"' or 'drools:ruleUnit="unit:org.kie.kogito.examples.CalculationService"' 
+    -->
+    <bpmn2:businessRuleTask id="ServiceTask_1" drools:ruleFlowGroup="unit:CalculationService" name="Calculate total" implementation="http://www.jboss.org/drools/rule">
```
mswiderski: what is the reason to replace serviceTask with businessRuleTask in this scenario?  if user needs to implement a service class then serviceTask sounds like better fit than business rule task

jbpm-quarkus-example/src/main/resources/org/kie/kogito/examples/person-rules.drl
evacchi: i think i got things mixed up :)

PR #77, opened 2019-08-22
Add Onboarding Kogito Operator test

sutaakar:


ricardozanini:
Is this test running on a real cluster? Cool.

Only thing that I could spot is that the onboarding service also has a label in the examples. And I didn’t see a check if the labels were applied in the services objects after KogitoApp creation. 

The objects creation seems fine :)

sutaakar:
good catch, I will add it there.
Yes, it is tested on real cluster.

sutaakar:
Minor enhancements here and there, service label test added.

ricardozanini (approved):

sutaakar:

mswiderski (approved):

PR #78, opened 2019-08-23
Implement Drools Quarkus example test using Kogito CLI

sutaakar:


ricardozanini:
Looks nice! The intention of this implementation is to run the operator via CLI and deploy the example application, right? Then I imagine that you verify the deploy with other test cases? Like, you can use CRs or CLI to deploy the app via the application and then the verification ends up being the same.

sutaakar:
yes, verification is done in test base classes (https://github.com/kiegroup/kogito-examples/blob/master/openshift-example-tests/src/test/java/org/kogito/examples/openshift/DroolsQuarkusExampleTestBase.java , https://github.com/kiegroup/kogito-examples/blob/master/openshift-example-tests/src/test/java/org/kogito/examples/openshift/OnboardingQuarkusExampleTestBase.java) and it is same for all approaches (manual with OC client, operator with manual CR, Kogito CLI).

ricardozanini (approved):

sutaakar:

mswiderski (approved):

PR #79, opened 2019-08-23
KOGITO-183: Explore annotation-based rule definition

evacchi:
related to https://github.com/kiegroup/kogito-runtimes/pull/130
currently this hacks DRL compilation by directing the Maven plugin to target/generated-sources/annotation, because an annotation processor is used. we may actually do reflection-based annotation processor from within the `kogito-codegen` framework


evacchi:
closing the example, let's just add a test case for now

PR #84, opened 2019-09-05
Switch onboarding deployment creations for onboarding test

sutaakar:
Create onboarding as a last deployment.
Previous the test request was sent to onboarding app even before HR app was completely initialized and its service available from onboarding app - caused test instabilities.
By building onboarding app as latest we ensure that once onboarding app is built and started the other apps are already available.


ricardozanini (approved):
LGTM

sutaakar:

PR #93, opened 2019-10-01
Removing DMN codegen to not clash with paymentDate endpoint

ricardozanini:
This will fix this WARN:

```
2019-10-01 14:55:24,767 WARN  [org.jbo.res.res.i18n] (executor-thread-1) RESTEASY002142: Multiple resource methods match request "POST /paymentDate". Selecting one. Matching methods: [public org.kie.kogito.dmn.rest.DMNResult http_58_47_47www_46trisotech_46com_47definitions_47_a412928f_4561f7_45428c_45a0bb_45150fda386899.PaymentDateResource.dmn(java.util.Map), public org.kie.kogito.examples.payroll.Payroll org.kie.kogito.examples.PaymentDateDMNEndpoint.calculatePaymentDate(org.kie.kogito.examples.payroll.Payroll)]
```

That could ends up with this ERROR:

```
2019-10-01 14:55:25,349 ERROR [io.und.req.io] (executor-thread-1) Exception handling request 1142bfd2-cfc1-4df1-a882-f0ffcfbe72fc-1 to /paymentDate: org.jboss.resteasy.spi.UnhandledException: java.lang.RuntimeException: Multiple model with the same name: paymentDate
```

### DIFF
--- a/onboarding-example/payroll/pom.xml
+++ b/onboarding-example/payroll/pom.xml
@@ -14,6 +14,9 @@
   <packaging>jar</packaging>
   <name>payroll</name>
   <description>Payroll related rules and processes for onboarding</description>
+  <properties>
+    <kogito.codegen.decisions>false</kogito.codegen.decisions>
+  </properties>
 
   <dependencyManagement>
     <dependencies>


ricardozanini:
@tarilabs you might want to take a look at this workaround. The auto generated code for DMN files clashed with the manually created endpoint for this demo.

mswiderski (approved):

tarilabs:
sorry, but not following --anyhow thanks for the fix !

mswiderski:
@tarilabs the endpoints that are generated have the same path like `/vacationDays` but then have different input (Map) and output (DMNResult) and these do not match expectations as they are not domain specific and expose data like messages, context etc that the client is not aware of and thus will fail.

tarilabs:
This demo is intended to NOT leverage endpoint generation, if in the src/main/java the user have created its own JAX-RS endpoint, isn't it ?

mswiderski:
this example has been written that way to fit the over all idea of domain specific services. But since the DMN endpoint generation was introduced it caused this example to fail on that assumption. And that is the root issue. It's part of the onboarding example and thus should work in complete setup. So intention is that it should leverage code generation but that should also fit the domain specific usage.

tarilabs:
sure, but if in this case the endpoint is generated manually, I would expect normal for the autogeneration to be disabled, isn't it ?

Also, evaluating a DMN is always a twofold exercise, you want back both the result and the error messages. And the spec requires to return both as it could happen you could only partially evaluate a model, and the rest to have failed with some message reported.
Do you have a concrete alternative example how I'm supposed to return both "result" and "error message" in a different way which is allegedly more aligned to this example then?

ricardozanini:
@tarilabs @mswiderski I'm still seeing this error:

```
Caused by: java.lang.RuntimeException: Multiple model with the same name: paymentDate
	at org.kie.kogito.dmn.DMNKogito.modelByName(DMNKogito.java:81)
	at org.kie.kogito.dmn.DMNKogito.evaluate(DMNKogito.java:86)
	at org.kie.kogito.examples.PaymentDateDMNEndpoint.calculatePaymentDate(PaymentDateDMNEndpoint.java:41)
```

I'll investigate a little bit further, but only happens on OpenShift with the service discovery. Might be related with the payload, I don't know yet. Do you guys have any clues?

tarilabs:
@ricardozanini can you share the payload you are using or better clarify to me the hypothetical statement : "That could ends up with this ERROR" so I have elements to reproduce on this side?

The DMN quarkus example is working fine, hence the issue is limited to this other demo.

ricardozanini:
@tarilabs I'm working on it. It's tough to get proper logs from the cluster.

mswiderski:
> sure, but if in this case the endpoint is generated manually, I would expect normal for the autogeneration to be disabled, isn't it ?

it should be if the generated endpoints are not suitable ... but was not and thus this fix

> Also, evaluating a DMN is always a twofold exercise, you want back both the result and the error messages. 

why to force users to parse the payload to know if there are errors or not? wouldn't it make more sense to simply return it via error code since this is error anyway... messages can also be included as headers or as payload in case of error state

> And the spec requires to return both as it could happen you could only partially evaluate a model, and the rest to have failed with some message reported.

so spec defines rest api that DMN can expose?

> Do you have a concrete alternative example how I'm supposed to return both "result" and "error message" in a different way which is allegedly more aligned to this example then?

see above

tarilabs:
> why to force users to parse the payload to know if there are errors or not? 

I believe user want to know if a decision resulted in null or another value, despite errors.

> wouldn't it make more sense to simply return it via error code since this is error anyway... 

error code is only 1, while I can have multiple errors

> messages can also be included as headers or as payload in case of error state

If I include in the payload you would need client side 2 different parsers, isn't this even more complicated for end-users?

> so spec defines rest api that DMN can expose?

unfortunately, not.
As I mentioned several times however the spec requires to return "results" and "error messages" if any.

Wouldn't make more sense to discuss this in a meeting?
You have a very strict requirements on how you want these payload to look like, and I'm just trying to understand how to fit the DMN-square-in-circled-hole but frankly I am not sure how I can accomodate your mandatory points?

ricardozanini:
please disregard this last commit, just put it there to get the logs in the OCP

tarilabs:
@ricardozanini the reason I was asking, is that I am failing to understand why the included tests do not exhibit the problem you are allegedly reporting. Hence I need a way to reproduce to understand if there is any fix really necessary on codegen (although so far it seems more a clash in requirement, since the endpoint are described manually, instead of relying on codegen)

ricardozanini:
@tarilabs here's the payload:

```
2019-10-01 17:43:41,256 DEBUG [org.kie.kog.clo.wor.ser.dis.BaseServiceDiscovery] (executor-thread-1) Result of services query: [{metadata={name=payroll-service, namespace=prometheus, selfLink=/api/v1/namespaces/prometheus/services/payroll-service, uid=26cdca38-e456-11e9-8e2d-0615e4bfa428, resourceVersion=13989215, generation=1, creationTimestamp=2019-10-01T14:17:08Z, labels={app=payroll-service, paymentDate=process, taxRate=process, vacationDays=process}, annotations={org.kie.kogito/managed-by=Kogito Operator, org.kie.kogito/operator-crd=KogitoApp}, ownerReferences=[{apiVersion=app.kiegroup.org/v1alpha1, kind=KogitoApp, name=payroll-service, uid=b0ce7d80-e454-11e9-ad11-02ff077e3f52, controller=true, blockOwnerDeletion=true}]}, spec={ports=[{name=http, protocol=TCP, port=8080, targetPort=8080}], selector={app=payroll-service}, clusterIP=172.30.3.130, type=ClusterIP, sessionAffinity=None}, status={loadBalancer={}}}]
2019-10-01 17:43:41,256 DEBUG [org.kie.kog.clo.wor.DiscoveredServiceWorkItemHandler] (executor-thread-1) Found endpoint for service paymentDate with location ServiceInfo [url=http://172.30.3.130:8080/paymentDate, headers={}]
2019-10-01 17:43:41,257 DEBUG [org.kie.kog.clo.wor.DiscoveredServiceWorkItemHandler] (executor-thread-1) Sending body {"employee":{"firstName":"Mark","lastName":"Test","personalId":"xxx-yy-yyyyzzz","birthDate":"1995-12-10T12:50:12.123+00:00","address":{"street":"any street 3","city":"Boston","zipCode":"10001","country":"US"}}}
2019-10-01 17:43:41,284 DEBUG [org.kie.kog.clo.wor.DiscoveredServiceWorkItemHandler] (executor-thread-1) Resonse code 500 and payload
2019-10-01 17:43:41,285 ERROR [org.jbp.wor.ins.imp.WorkflowProcessInstanceImpl] (executor-thread-1) Unexpected error (id e4592530-4952-4654-98fc-90811c678666) while executing node Calculate payment date in process instance c78942a7-e390-4816-9f45-a9a51a3935c0: org.jbpm.workflow.instance.WorkflowRuntimeException: [onboarding.setupPayroll:c78942a7-e390-4816-9f45-a9a51a3935c0 - Calculate payment date:1] -- Unsuccessful response from service Internal Server Error (code 500)
```

The entire log can be seen at:
https://pastebin.com/S0xYqXw3 (onboarding)

The payroll service error: 
https://pastebin.com/QzeYMucu (payroll)

Thanks!

mswiderski:
> error code is only 1, while I can have multiple errors

You can just use the code to indicate that there is an error while the payload eill come with all the details

> If I include in the payload you would need client side 2 different parsers, isn't this even more complicated for end-users?

I would say that this is standard practice to have different data returned upon successful response or an error so in my opinion this is what would be expected

> unfortunately, not. As I mentioned several times however the spec requires to return "results" and "error messages" if 

So why do you want to expose spec details to consumers of the service? In other words why consumers need to even know or care that behind is a DMN engine at all? This is against on of the main themes of kogito - ergo domain - that means no leaking abstraction of the t3chnology behind it

> Wouldn't make more sense to discuss this in a meeting?

Feel free to setup a meeting

> You have a very strict requirements on how you want these payload to look like, and I'm just trying to understand how to fit the DMN-square-in-circled-hole but frankly I am not sure how I can accomodate your mandatory points?

Again, this is nit about me but main principles of kogito. As mentioned above and to emphasize kogito is more than running rules, processes and decision on quarkus

tarilabs:
@ricardozanini I am getting a more understanding about the different stages on this PR, but I get a very specific question to try to understand fully the original problem.
Let me be very precise.
Before this PR, you had the WARN, this I can understand why.
Still, and again before the PR, was there a way to induce the error 

```
equest 1142bfd2-cfc1-4df1-a882-f0ffcfbe72fc-1 to /paymentDate: org.jboss.resteasy.spi.UnhandledException: java.lang.RuntimeException: Multiple model with the same name: paymentDate
```

? 
Because I was not able to reproduce that error specifically just yet.

tarilabs:
Tried to ping via chat but unsuccessfully, so I will try to summarize my understanding here.

I believe from this comment https://github.com/kiegroup/kogito-examples/pull/93#issue-323336510 : 
I can agree about the WARN statement.
I am not able to understand nor to reproduce the mentioned _"could ends up with this ERROR"_ in any way.
Hence I believe from mentioned comment, the first statement is correct, the second statement is not reproducible.

I believe from this comment https://github.com/kiegroup/kogito-examples/pull/93#issuecomment-537096306 : 
theoretically valid, but the error reported by Ricardo does not point to that kind of error. As it would be a type/cast error. Hence it's a separate discussion.

I believe from this comment https://github.com/kiegroup/kogito-examples/pull/93#issuecomment-537104679 : 
judging by the stacktrace, it cannot be a problem of the payload, since it's clearly going through the bespoke REST Endpoint.
The only other reason that reported RuntimeException can be thrown, is that when the java code initialised, the runtime did not scan the `.dmn` resources, so the message says "multiple model with the same name" but I am starting to doubt for some reason **zero** models were originally detected under that name.

> only happens on OpenShift with the service discovery

I believe the answer might be hidden somewhere there ^.
Can I get some more details about this environment, please?
Specifically, I would need to check when the project was deployed, the `.dmn` resources are indeed available inside the .JAR.
Is there a way to check that ?

I am available via chat if more suitable. Please let me know.

ricardozanini:
@tarilabs I'm sorry but I wasn't able to chat at that time. For this PR please consider only the `WARN` thing discussed here that this patch resolves.

For the ERROR part, I guess we have to open a discussion in another channel. Is pretty late here and I should get back to you tomorrow morning (afternoon for you)

Many thanks for taking a look into this. Tomorrow we can get the image/jar file and inspect them.

See ya

tarilabs:
I believe I was able to reproduce the problem, locally.

You can merge this PR to avoid the "WARN" originally reported here: https://github.com/kiegroup/kogito-examples/pull/93#issue-323336510

However, as discussed that won't solve the "error multiple model" when on OCP.
I am working to fully isolate the problem and will eventually provide a solution separately.

ricardozanini:
@mswiderski I guess we can merge this. The error is being handled here: https://github.com/kiegroup/kogito-examples/pull/94

PR #95, opened 2019-10-04
move department service of onboarding hr to an query based automatica…

mariofusco:
…lly generated rest endpoint


evacchi:
outdated

PR #97, opened 2019-10-17
Update to Stunner!

porcelli:
Update the XML format to Stunner, so it can look good in VSCode


evacchi:
is this PR outdated?

porcelli:
i think so, i think @inodeman has been working on updating stuff around... right?

porcelli:
ops.. I think this might be still valid :o

we need convert the current file to stunner.

domhanak:
Closing as outdated

PR #98, opened 2019-10-19
[KOGITO-268] allow to declaratively use a SessionPool for KieSesssion…

mariofusco:
…s used by rule units


drools-quarkus-unit-example/src/main/resources/META-INF/kmodule.xml
```diff
@@ -1,9 +1,2 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<kmodule xmlns="http://www.drools.org/xsd/kmodule">
-    <kbase name="alertingKB" packages="org.kie.kogito.rules.alerting">
-        <ksession name="alertingKS"/>
-    </kbase>
-    <kbase name="queriesKB" packages="org.kie.kogito.queries">
-        <ksession name="queriesKS"/>
-    </kbase>
-</kmodule>
+<kmodule xmlns="http://www.drools.org/xsd/kmodule"/>
```
evacchi, line 2: FYI: it shouldn't be necessary to add this anymore. a Default kmodule model is created in memory if missing

PR #129, opened 2020-02-04
KOGITO-1018 - Enable JUnit reporting on kogito-examples

MarianMacik:
I've noticed that JUnit reporting was turned off from the old days when there were no tests in the repository.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -53,8 +53,14 @@ pipeline {
             steps {
                 script {
                     maven.runMavenWithSubmarineSettings('clean install', false)
-                    // Don't run with tests so far, see: https://github.com/quarkusio/quarkus/issues/6885
-                    maven.runMavenWithSubmarineSettings('clean install -Ppersistence', true)
+                }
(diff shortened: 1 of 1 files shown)


radtriste (approved):

ricardozanini (approved):

ricardozanini (approved):

Jenkinsfile
```diff
@@ -71,8 +77,7 @@ pipeline {
             }
         }
         always {
-            // Currently there are no tests in submarine-examples
-            //junit '**/target/surefire-reports/**/*.xml'
+            junit '**/target/surefire-reports/**/*.xml'
```
radtriste, line 80: should we use `allowEmptyResults` here ?
cf https://jenkins.io/doc/pipeline/steps/junit/

MarianMacik: Nope, this was because by a previous PR I actually cleaned the results from run without persistence :) But in other repositories it didn't fail the build as there were tests from other repositories. I have to fix it in others as well. It's just 70 tests or so, but still.

PR #130, opened 2020-02-07
[KOGITO-1023] consolidate kogito-bom and kogito-runtimes repos

mariofusco:


pom.xml
```diff
@@ -66,50 +66,43 @@
```
MarianMacik: Still inheriting from kogito-bom-parent.

MarianMacik (changes requested):

MarianMacik:
Also please delete `kogito-bom` stage from `Jenkinsfile`:
https://github.com/kiegroup/kogito-examples/blob/83586f0107ddc9c89d5916a6dc76d9b944678a4d/Jenkinsfile#L21-L30

evacchi (approved):
I think it should be ok now

PR #135, opened 2020-02-18
[KOGITO-858] Add a Decision Table RuleUnit example

tkobayas:
decisiontable-quarkus-example and decisiontable-springboot-example.

Please review, thanks!


decisiontable-quarkus-example/src/test/java/org/kie/kogito/QueryTest.java
```diff
@@ -0,0 +1,84 @@
...
+import org.kie.kogito.queries.LoanUnitQueryFindApprovedEndpoint;
+import org.kie.kogito.queries.LoanUnitQueryFindNotApprovedIdAndAmountEndpoint;
+import org.kie.kogito.queries.LoanUnitRuleUnit;
+
+import static java.util.Arrays.asList;
+import static java.util.stream.Collectors.toList;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class QueryTest {
```
evacchi: it's probably better for now to remove these tests against the programmatic APIs, as 1) they duplicate the http tests 2) they test an API that is still not stable nor public 3) they pose some problems when we use the quarkus extension instead of the maven plugin (basically they break the build because unfortunately the extension runs *after* the tests, in the package phase...)

decisiontable-springboot-example/src/test/java/org/kie/kogito/QueryTest.java
```diff
@@ -0,0 +1,84 @@
...
+import org.kie.kogito.queries.LoanUnitQueryFindApprovedEndpoint;
+import org.kie.kogito.queries.LoanUnitQueryFindNotApprovedIdAndAmountEndpoint;
+import org.kie.kogito.queries.LoanUnitRuleUnit;
+
+import static java.util.Arrays.asList;
+import static java.util.stream.Collectors.toList;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class QueryTest {
```
evacchi: let's remove it from here too, to avoid confusion

evacchi (commented):
looks good, but let's wait for #130 to be merged, this way we can already update the quarkus example to use the quarkus extension instead of the maven plugin

decisiontable-quarkus-example/src/test/java/org/kie/kogito/QueryTest.java
MarianMacik: So that is the reason why we use `kogito-maven-plugin` also with Quarkus examples instead of Kogito Quarkus extension?

decisiontable-springboot-example/src/test/java/org/kie/kogito/QueryTest.java
MarianMacik: +1 for consistency

MarianMacik (commented):
Looks good after addressing Edoardo's comments.

decisiontable-quarkus-example/src/test/java/org/kie/kogito/QueryTest.java
evacchi: no, that's a side effect. 

the reason why we use the maven plugin is it was too hard to use the latest codegen (SNAPSHOT) because the quarkus bom controlled versioning of the extension (and therefore of the runtimes). If we wanted to use the extension, we would have to redeclare all dependencies explicitly in all poms. Now that we can control it, we can provide our own "kogito-quarkus-bom" pointing to SNAPSHOT

tkobayas: Removed QueryTest. Thanks!

decisiontable-springboot-example/src/test/java/org/kie/kogito/QueryTest.java
tkobayas: Removed QueryTest. Thanks!

tkobayas:
Thank you for reviewing!

I have updated the PR:
- Use quarkus extension instead of kogito-maven-plugin
- Removed QueryTest from decisiontable-quarkus-example and decisiontable-springboot-example

decisiontable-quarkus-example/pom.xml
```diff
@@ -0,0 +1,137 @@
...
+          <execution>
+            <goals>
+              <goal>build</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>io.openapitools.swagger</groupId>
+        <artifactId>swagger-maven-plugin</artifactId>
```
evacchi: you should be able to get rid of this plugin and use `quarkus-smallrye-openapi` instead :)

evacchi (approved):
one last change and we should be good to go

decisiontable-quarkus-example/pom.xml
tkobayas: Good to know. Thanks!

tkobayas:
Update the PR:
- Removed swagger-maven-plugin from decisiontable-quarkus-example

MarianMacik (approved):

PR #137, opened 2020-02-27
KOGITO-1297: Cannot package kogito-examples because of kogito-codegen bug

evacchi:
I can't reproduce the same bug exactly, however, there were a few build errors, that I am fixing with this PR


radtriste (approved):
```
$ rm -rf ~/.m2/repository/org/kie/kogito/
$ mvn clean package
```

working

PR #154, opened 2020-03-18
Use project version as default Kogito version

cristianonicolai:


cristianonicolai:
@evacchi this is needed so when we do a tag, project is set up correctly to whatever version is released.

evacchi (approved):

PR #157, opened 2020-03-20
Quarkus version of the spring-boot flight example.

Christopher-Chianelli:
TODO:
- Update README
- Fix process test


jbpm-optaplanner-quarkus-example/.gitignore
```diff
@@ -0,0 +1,32 @@
...
+# Test info
+/settings*.xml
+/lib-jdbc/
+*.db
+*.tlog
+
+# modules that don't exist in this branch
+/jbpm-human-task-war/
+/jbpm-bam/
+/jbpm-gwt/
```
ge0ffrey: Let's remove these 4 lines.

jbpm-optaplanner-quarkus-example/.gitignore
```diff
@@ -0,0 +1,32 @@
...
+
+# files used for external db testing
+jdbc_driver.jar
+db-settings.xml
```
ge0ffrey: Let's remove these 3 lines

jbpm-optaplanner-quarkus-example/.gitignore
```diff
@@ -0,0 +1,32 @@
...
+.DS_Store
+
+# Original jbpm ignores
+*~
+
+# Test info
+/settings*.xml
+/lib-jdbc/
+*.db
+*.tlog
```
ge0ffrey: Let's remove these 5 lines

ge0ffrey: Actually, non of the other examples in this repo have a .gitignore file, let's remove this file entirely.

There are basically 2 ways to deal with .gitignore:
A) a single root .gitignore file that says to exclude all "target" and ".idea" directories, in any subdirectory.
B) a .gitingore file per module tot exclude "/target" in this directory only.

Pro/con: B) is more work, but A) is more error prone (think java package org.foo.target.goal)

This repository is employing tactic A), so this PR should adhere to that and not have a .gitignore file here.

jbpm-optaplanner-quarkus-example/README.md
```diff
@@ -0,0 +1,341 @@
+# jBPM + OptaPlanner + SpringBoot example
```
ge0ffrey: do a "find in directory" for "spring" :)

jbpm-optaplanner-quarkus-example/README.md
```diff
@@ -0,0 +1,341 @@
...
+
+## Installing and Running
+
+```
+mvn clean package spring-boot:run    
+```
+
+## Swagger documentation
+
+You can take a look at the [swagger definition](http://localhost:8080/swagger.json) - automatically generated and included in this service - to determine all available operations exposed by this service.  For easy readability you can visualize the swagger definition file using a swagger UI like for example available [here](https://editor.swagger.io). In addition, various clients to interact with this service can be easily generated using this swagger definition.
```
ge0ffrey: Does the quarkus version support swagger too?

"You can take a look at ..." -> "Take a look at ..."

jbpm-optaplanner-quarkus-example/pom.xml
```diff
@@ -0,0 +1,236 @@
...
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-bom</artifactId>
+        <version>${kogito.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+      <dependency>
+        <groupId>javax</groupId>
+        <artifactId>javaee-api</artifactId>
+        <version>8.0</version>
```
ge0ffrey: This dependency is wierd.
In a quarkus world, depending on the quakus extensions alone should be enough. Can we remove this dependency?

jbpm-optaplanner-quarkus-example/pom.xml
```diff
@@ -0,0 +1,236 @@
...
+        <groupId>org.optaplanner</groupId>
+        <artifactId>optaplanner-bom</artifactId>
+        <version>${version.org.optaplanner}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+      <dependency>
+        <groupId>org.webjars</groupId>
+        <artifactId>webjars-locator</artifactId>
+        <version>0.37</version>
```
ge0ffrey: The optaplanner-quickstart for quarkus doesn't have this dependency, because we couldn't get that to work.
1) Do a diff between this pom and the optaplanner-quickstart pom to verify there is nothing in here that you don't need
2) If you can get the locator to work, such as the js file urls don't have a version number, that would be great (out of scope for this PR to get it merged asap I would argue) and then we'd like to consider that for the optaplanner-quickstart example too :) I do suspect that's just a missing feature of quarkus...

jbpm-optaplanner-quarkus-example/pom.xml
```diff
@@ -0,0 +1,236 @@
...
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-resteasy-jackson</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-optaplanner-jackson</artifactId>
+    </dependency>
```
ge0ffrey: Move this quarkus-optaplanner-jackson directly under the optaplanner one, it's easier to read. (Code is read 10 times more than it is written, so readability matters).

jbpm-optaplanner-quarkus-example/pom.xml
```diff
@@ -0,0 +1,236 @@
...
+
+    <dependency>
+      <groupId>javax</groupId>
+      <artifactId>javaee-api</artifactId>
+      <scope>provided</scope>
+    </dependency>
```
ge0ffrey: Can we remove this one?

jbpm-optaplanner-quarkus-example/pom.xml
```diff
@@ -0,0 +1,236 @@
...
+              <goal>build</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <!-- <plugin>
+        <groupId>org.kie.kogito</groupId>
+        <artifactId>kogito-maven-plugin</artifactId>
+        <extensions>true</extensions>
+      </plugin> -->
```
ge0ffrey: Not needed, remove these 4 lines.

Btw: any comment here, please also check the springboot example for that and if it applies there too (and it's not exceptionally different like the webjar locator which is supported there) then adjust it there too please. So if this commented out code is there, remove it please there too.

jbpm-optaplanner-quarkus-example/pom.xml
```diff
@@ -0,0 +1,236 @@
...
+      <plugin>
+        <groupId>io.openapitools.swagger</groupId>
+        <artifactId>swagger-maven-plugin</artifactId>
```
ge0ffrey: Does this work with Quarkus? :)

jbpm-optaplanner-quarkus-example/src/main/java/org/kie/kogito/examples/domain/Seat.java
```diff
@@ -0,0 +1,72 @@
...
+                    this.emergencyExitRow == other.emergencyExitRow;
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        // Name is computed based on row and column and thus does not need to be included in the hash.
+        return Objects.hash(row, column, seatType, emergencyExitRow);
+    }
```
ge0ffrey: Are business intelligent equals/hashCode methods needed for the quarkus and spring boot example to work?
Less is more: if it works without these methods on the domain objects, please remove them, otherwise users think they are crucial.

jbpm-optaplanner-quarkus-example/src/main/java/org/kie/kogito/examples/service/FlightSeatingSolveService.java
```diff
@@ -0,0 +1,46 @@
...
+        new Thread(() -> {
+            try {
+                Flight finalBestSolution = solverJob.getFinalBestSolution();
+                process.instances().findById(id).ifPresent(pi -> {
+                    pi.send(Sig.of("solvingTerminated", finalBestSolution));
+                });
+            } catch (InterruptedException | ExecutionException e) {
+                e.printStackTrace();
+            }
+        }).start();
```
ge0ffrey: Oh yes, I need to prioritize PLANNER-1868 so there's no need to do "new Thread". Let's merge this as is and feel free to remind me about 1868 if I haven't fixed that next week.

jbpm-optaplanner-quarkus-example/src/main/java/org/kie/kogito/examples/solver/FlightSeatingConstraintProvider.java
```diff
@@ -0,0 +1,62 @@
...
+        return factory.from(Passenger.class).groupBy(ConstraintCollectors.sum(p -> p.getSeat().getColumn()))
+                .join(factory.from(Passenger.class).groupBy(ConstraintCollectors.sum(p -> p.getSeat().getRow())))
+                .join(factory.from(Passenger.class).groupBy(ConstraintCollectors.count()))
+                .join(FlightInfo.class)
+                .penalize("Plane Balance", HardSoftScore.ONE_SOFT, (totalX, totalY, passengerCount, flightInfo) -> {
+                    double localTotalX = totalX - passengerCount * ((flightInfo.getSeatColumnSize() / 2.0) - 0.5);
+                    double localTotalY = totalY - passengerCount * ((flightInfo.getSeatRowSize() / 2.0) - 0.5);
+                    double averageX = localTotalX / passengerCount;
+                    double averageY = localTotalY / passengerCount;
+                    return (int) Math.round(Point2D.distance(0, 0, averageX, averageY) * passengerCount * 100);
```
ge0ffrey: we might want to do this in "milis" or maybe not. Something to discuss, please bring this subject up in our next sync-up so I can explain the "milis" approach and we can determine if it's needed here.

jbpm-optaplanner-quarkus-example/src/main/resources/META-INF/resources/index.html
```diff
@@ -0,0 +1,171 @@
+<html lang="en">
+<head>
+    <meta content="text/html; charset=UTF-8" http-equiv="Content-Type">
+    <meta content="width=device-width, initial-scale=1" name="viewport">
+    <title>OptaPlanner Spring Boot example</title>
```
ge0ffrey: ctrl-shift-f is your friend :)

pom.xml
```diff
@@ -129,6 +129,7 @@
     <module>jbpm-quarkus-example</module>
     <module>jbpm-springboot-example</module>
     <module>jbpm-optaplanner-springboot-example</module>
+    <module>jbpm-optaplanner-quarkus-example</module>
```
ge0ffrey: Quarkus First :) Please move this line one line up.

jbpm-optaplanner-quarkus-example/pom.xml
Christopher-Chianelli: Just tested, we can; unrelated, IDE cannot find Optaplanner, talks about missing dependency version for the OptaPlanner extensions; it compiles and runs fine though.

jbpm-optaplanner-quarkus-example/pom.xml
Christopher-Chianelli: Nope, got a JSON with only title and version after changing the location. Removing.

jbpm-optaplanner-quarkus-example/src/main/java/org/kie/kogito/examples/domain/Seat.java
Christopher-Chianelli: Not needed; @triceo advocated for it since `Seat` is immutable. Curious, how does OptaPlanner handles `equals` but not `is` (that is, `x.equals(y)` but not `x == y`)? If I pin a passenger to a Seat that is not a Seat on the Flight, but rather a Seat equal to a Seat on the Flight, will there be no adverse side effects (the Seat won't be in a Problem Fact Collection)?

Christopher-Chianelli: (If it handled well,
```
            String[] seatLoc = passenger.getSeat().split(";");
            int row = Integer.parseInt(seatLoc[0]);
            int col = Integer.parseInt(seatLoc[1]);
            seat = flight.getSeatList().stream().filter(s -> s.getRow() == row && s.getColumn() == col).findAny().get();
```
can be removed and we can just access the Seat directly from the DTO.

jbpm-optaplanner-quarkus-example/README.md
Christopher-Chianelli: `jbpm-quarkus-example/README.md ` also mentions Swagger, but doesn't have Swagger in the `pom.xml` (nor anything swagger related in resources)

jbpm-optaplanner-quarkus-example/pom.xml
Christopher-Chianelli: Previous reasoning was to have all jackson modules grouped together (there are three: quarkus-jackson, resteasy-jackson, and optaplanner-jackson); moved it under optaplanner.

jbpm-optaplanner-quarkus-example/pom.xml
Christopher-Chianelli: Not there; this code is needed for spring boot, but not for Quarkus.

jbpm-optaplanner-quarkus-example/pom.xml
Christopher-Chianelli: I doubt we can get the locator to work due to how webjars work in Quarkus. Quarkus put webjars in `target/lib` and reference them in the `META-INF/MANIFEST.MF`'s `Class-Path`.

Meanwhile, Spring Boot put the webjars directly in `BOOT-INF/lib/`. 

From what I see in `Spring`, `Spring` doesn't actually use the `webjar-locator`, but rather use its existance as a trigger to server versionless resources instead of versioned resources via https://github.com/spring-projects/spring-boot/blob/7f509bf84e9b51416ce65a777a3b5b2521921a6f/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/WebMvcAutoConfiguration.java#L299-L317

Christopher-Chianelli: (the condition: https://github.com/spring-projects/spring-boot/blob/aef92b9295f62d008faa9ab79905a474bf3496f3/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/OnEnabledResourceChainCondition.java)

and the customizer:

https://github.com/spring-projects/spring-boot/blob/7f509bf84e9b51416ce65a777a3b5b2521921a6f/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/WebMvcAutoConfiguration.java#L519-L557

Christopher-Chianelli: My guess is we can either add a new quarkus extension for this or add the functionality to the resteasy extension.

jbpm-optaplanner-quarkus-example/src/main/java/org/kie/kogito/examples/domain/Seat.java
ge0ffrey: OptaPlanner expects that given the following input:
```
- Rooms [A, B]
- Lessons [ {Math, room: A}, {French, room: A}]
```
that all those A instances are the same instance. (That's btw that in JPA the rooms and lessons need to be fetched in the same transaction).

Then, the output of OptaPlanner also respects this philosophy. Basically, a solution will never have 2 room A instances. Or two seat instances. (Even if planning cloning happens that still holds true) So default Object.hashCode/equals() suffices. Less code is better than more code.

jbpm-optaplanner-quarkus-example/pom.xml
ge0ffrey: Great work on building that new quarkus extension. This PR can't wait for the quarkus 1.4 release (which hopefully has the webjar extension), so +1 to workaround it for now. Do create a jira to remove that workaround in this example once quarkus 1.4 is released.

ge0ffrey (approved):

ricardozanini:
Is this still a draft?

Christopher-Chianelli:
@ricardozanini I still need to implement a test, so yes (I needed to work on an urgent task this sprint, leading most of my PR's abandoned for the past 3 weeks, going back to finish them now)

Christopher-Chianelli:
@ricardozanini added test for both jvm and native mode; removed draft status

ge0ffrey:
jenkins retest this

process-optaplanner-quarkus/README.md
```diff
@@ -0,0 +1,358 @@
...
+and custom Java service, a new service is generated that exposes REST operations to
+create new flights (following the steps as defined in the main and sub-process)
+and add passengers to flights.
+
+A UI is included in this example, which can be accessed on `localhost:8080`
+when the application is running.
+
+## Installing and Running
+
+### Prerequisites
```
evacchi, line 27: align prerequisites to the rest of the repo (java 11, maven 3.6.x)

evacchi (commented):
I think it looks good, but I think you'll have to verify the Quarkus dependency and/or rebase on master, since the build is currently red.

jbpm-optaplanner-quarkus-example/README.md
evacchi: you have to add the OpenAPI Quarkus extension. You can also set it to always enabled (otherwise it's only available in devmode)

process-optaplanner-quarkus/README.md
```diff
@@ -0,0 +1,364 @@
+# jBPM + OptaPlanner + Quarkus example
```
ricardozanini:
```suggestion
# Process + OptaPlanner + Quarkus example
```

process-optaplanner-quarkus/pom.xml
```diff
@@ -0,0 +1,196 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.kie.kogito</groupId>
+    <artifactId>kogito-examples</artifactId>
+    <version>8.0.0-SNAPSHOT</version>
+  </parent>
+  <groupId>org.kie.kogito.examples</groupId>
+  <artifactId>process-optaplanner-quarkus</artifactId>
+
+  <name>Kogito Example :: jBPM, OptaPlanner and Quarkus</name>
```
ricardozanini:
```suggestion
  <name>Kogito Example :: Process, OptaPlanner and Quarkus</name>
```

Please also look up after this word in the README and any docs in this example.

ricardozanini (changes requested):
Hi @Christopher-Chianelli !
Thanks for this example, but please consider also adding an example for the operator deployment on `operator` directory. For reference, see the OptaPlanner springboot example.

Christopher-Chianelli:
@ricardozanini Done (I think; don't know how to test locally)

ricardozanini (approved):

evacchi (approved):
if anything from others have been addressed this is good for me

PR #164, opened 2020-03-25
KOGITO-1555: Make sure Quarkus snapshot works with examples and runtimes

evacchi:
remove unnecessary dependency on 1.0.0.Alpha1

### DIFF
--- a/onboarding-example/hr/pom.xml
+++ b/onboarding-example/hr/pom.xml
@@ -72,6 +72,7 @@
           <plugin>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-maven-plugin</artifactId>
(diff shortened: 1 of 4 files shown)


onboarding-example/hr/pom.xml
```diff
@@ -72,6 +72,7 @@
           <plugin>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-maven-plugin</artifactId>
+            <version>${version.io.quarkus}</version>
```
cristianonicolai: @evacchi can we move this into the pluginmanagement section from the root pom? I tried to centralized all versions there.

cristianonicolai (approved):

PR #166, opened 2020-03-27
[KOGITO-1589] Unify OpenAPI/Swagger configuration and READMEs

jstastny-cz:
* I've configured openapi/swagger in all examples where it wasn't but there were non-trivial request examples in readme. For that matter I didn't touch helloworld examples - they show just a simple curl request.
* Unified openapi related sections in docs (different configuration for quarkus vs springboot results in different web context of the openapi definition or presence/absence of Swagger UI)
* In some cases (optaplanner only AFAIK) the location of swagger.json is different from others - caused by different setup in application.properties file.


timer-quarkus-example/src/main/resources/application.properties
```diff
@@ -0,0 +1 @@
+quarkus.swagger-ui.enable=false
```
evacchi: you can remove this entirely as this is the default. But I think we may actually want to set it to true to all of them; otherwise it is only available in devmode. WDYT?

evacchi:
overall looks good

krisv:
@jstastny-cz I believe the openapi module is still missing in jbpm-quarkus-example?  Could this be added in this PR (I believe it was removed unintendedly 2w ago)?

timer-quarkus-example/src/main/resources/application.properties
jstastny-cz: Default in development mode is true IMO [1]. Here I was not sure if it makes sense to add. If it does, I'll remove the disabling here.

[1] https://quarkus.io/guides/openapi-swaggerui#quarkus-swaggerui_quarkus.swagger-ui.enable

jstastny-cz: For enabling in native there is another property, documented at the same page. If we want to use the Swagger UI when deployed in OpenShift then it sounds like something we need to set, right?

jstastny-cz:
@krisv yes, it should have been here. I'll add it for sure.

ricardozanini (approved):

jstastny-cz:
@evacchi I've added the Swagger UI in Native mode as a separate commit..

evacchi (approved):

sterobin (approved):
@jstastny-cz , this is a welcome addition. Approved, but I did notice that some example apps mention the predefined Swagger UI available in Quarkus dev mode and native mode, while others don't. I assume this is because those ones don't support the Swagger UI. If that's not the case, and indeed all examples support the same UI, we should mention that in all applicable examples.

jstastny-cz:
> I did notice that some example apps mention the predefined Swagger UI available in Quarkus dev mode and native mode, while others don't

@sterobin currently only Quarkus based examples do bundle the Swagger UI, so only examples built on Quarkus should have that info in README.md. With one exception - the quarkus helloworld projects ... I didn't put it there cause I considered the config too complex for a helloworld project.
If there's something not adhering to this rule just let me know.

PR #172, opened 2020-04-01
Review Jenkinsfile for building stable branch

cristianonicolai:
This should only rely on published artifacts

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,8 +5,8 @@ pipeline {
         label 'kie-rhel7 && kie-mem16g'
     }
     tools {
-        maven 'kie-maven-3.6.0'
-        jdk 'kie-jdk1.8'
+        maven 'kie-maven-3.6.2'
+        jdk 'kie-jdk11'
     }
     options {
         buildDiscarder logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '', numToKeepStr: '10')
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -18,26 +18,6 @@ pipeline {
                 sh 'printenv'
             }
         }
-        stage('Build kogito-runtimes') {
-            steps {
-                dir("kogito-runtimes") {
-                    script {
-                        githubscm.checkoutIfExists('kogito-runtimes', "$CHANGE_AUTHOR", "$CHANGE_BRANCH", 'kiegroup', "$CHANGE_TARGET")
```
evacchi: maybe we can fix without removing by putting an if in front of this: `if (CHANGE_BRANCH != 'master' && CHANGE_TARGET != 'master') ` same down below

evacchi (changes requested):
good point! although as mentioned in the other thread, I think of the stable branch as a moving pointer so that never diverges from master. WDYT?

Jenkinsfile
cristianonicolai: I think is actually `if($CHANGE_TARGET != "stable")`, changes targetting any other branch should still work with full build.
Will test PR here, if it works, I will submit to master and others branches...

cristianonicolai:
> good point! although as mentioned in the other thread, I think of the stable branch as a moving pointer so that never diverges from master. WDYT?

true, that should make things easier to keep branches in sync.

PR #177, opened 2020-04-01
KOGITO-1544: Move kogito-travel-agency-tutorial to kogito-examples

evacchi:
I haven't added this to the main pom.xml yet. We can do this here or in a next PR. 
I would go with starting with merging and then do further changes later.
We are bringing in only the old '06'. But commit 3254a96 can be always restored if we want to bring back 01. We can also bring back 02,03,04 with commit 66cccd7 

**[PLEASE DON'T SQUASH](https://www.youtube.com/watch?v=YIAnkrPgTvY)**


evacchi:
sorry this should include 08 not 06. closing for now. will reopen

PR #179, opened 2020-04-03
[KOGITO-1674] removed testcontainers, fixed integration test phase in oidc examples

jstastny-cz:


evacchi (commented):
ok I don't dislike this change in general (invoking docker/podman instead of relying on test containers) although personally I would have limited to just move tests to integration. I would wait for others to have a say on this, as I am not an expert/don't know best practices in this case. The only downside I see, this probably won't work in IDEs, and I suppose test containers will.

tiagodolphine:
Yeah, I have the same opinion as @evacchi and can't we move the tests that depend on external services to integration-test, in this way the user can run this kind of test if he wants/have the proper requirements for it, but it is not the default behavior. Just my opinion, not against any decision, by the way, just open the discussion.

ricardozanini:
We definitely should move those tests to `IT`. Maybe users won't even have docker/podman installed locally (yeah, happens).

cristianonicolai:
@jstastny-cz seems overall consensus to just make the necessary changes to have tests in the integration-test phase. I think for now that would address the main issue, users not able to package the app without docker. We probably need to discuss how we want to address the testcontainer issue long term.

jstastny-cz:
@evacchi @tiagodolphine @ricardozanini 
moving the test to verify phase is one thing. The other I am trying to solve here is that testcontainers rely on local env too closely, it a) supports Docker only, b) requires SELinux exception (to access docker.sock) to be set. I don't think those are necessary prerequisites.
Moreover there are some enterprise linux systems that have podman only, no docker anymore :-)

jstastny-cz:
For the IDE experience ... I would argue that during development person should run the keycloak instance externally to save time anyway. In this regards the shift towards the maven plugin is fitting better than running it from Java. This way I can just add a maven property to skip the keycloak startup on one hand, or even to keep keycloak running (so that one can independently run tests from IDE against the instance) on the other.

bsideup:
One can "keep it running" with Testcontainers and the reusable containers feature (currently in preview):
https://github.com/testcontainers/testcontainers-java/pull/1781

jstastny-cz:
@evacchi @tiagodolphine @cristianonicolai I think there are two questions, one short-term, second long-term.
1. We need to be able to build examples anywhere, that calls for podman support. How can we do that in following days? I'd say this PR is the easiest path now (I personally can't run the testcontainers locally.)
2. Eventually, given that majority of testing is done upstream, we will need to able to build everything (including all the testing) using podman too. Will the testcontainers support arrive in time/at all?

Still, there's the issue with SELinux exception for accessing docker.sock needed. (Unless you disable ryuk container in testcontainers which leaves you with non-ideal env cleanup AFAIK.)

bsideup:
> Will the testcontainers support arrive in time/at all?

The question is more like: Will Podman ship the (promised) Docker API layer in time/at all :)

jstastny-cz:
> > Will the testcontainers support arrive in time/at all?
> 
> The question is more like: Will Podman ship the (promised) Docker API layer in time/at all :)

@bsideup
:-) fair enough, but not much changes here for our project. Though is that assumption correct? That podman support is quite unpredictable with regards to timing?

bsideup:
@jstastny-cz there was a spike but we haven't heard back from the Podman team what's the status. I assume you have a better communication channel with them and can ask directly ;)

jstastny-cz:
Closing in favour of #198

PR #182, opened 2020-04-03
KOGITO-1696: Set quarkus default ports to avoid conflicts

jomarko:
Set quarkus default ports for particular modules to avoid conflicts when running all modules
- hr : 8081
- payroll : 8082
- onboarding : 8080

For more details see https://issues.redhat.com/browse/KOGITO-1696

### DIFF
--- a/onboarding-example/hr/pom.xml
+++ b/onboarding-example/hr/pom.xml
@@ -49,10 +49,20 @@
     <plugins>
       <plugin>
         <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <systemProperties>
+            <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+            <quarkus.http.test-port>8081</quarkus.http.test-port>
+          </systemProperties>
+        </configuration>
       </plugin>
       <plugin>
         <groupId>io.quarkus</groupId>
         <artifactId>quarkus-maven-plugin</artifactId>
+        <configuration>
+          <debug>false</debug>
+          <jvmArgs>-Dquarkus.http.port=8081 -Dquarkus.http.host=localhost</jvmArgs>
+        </configuration>
         <executions>
           <execution>
             <goals>
--- a/onboarding-example/hr/pom.xml
+++ b/onboarding-example/hr/pom.xml
@@ -52,7 +52,6 @@
         <configuration>
           <systemProperties>
             <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
-            <quarkus.http.test-port>8081</quarkus.http.test-port>
           </systemProperties>
         </configuration>
       </plugin>
@@ -61,7 +60,6 @@
         <artifactId>quarkus-maven-plugin</artifactId>
         <configuration>
           <debug>false</debug>
-          <jvmArgs>-Dquarkus.http.port=8081</jvmArgs>
         </configuration>
         <executions>
           <execution>
--- a/onboarding-example/hr/pom.xml
+++ b/onboarding-example/hr/pom.xml
@@ -61,7 +61,7 @@
         <artifactId>quarkus-maven-plugin</artifactId>
         <configuration>
           <debug>false</debug>
-          <jvmArgs>-Dquarkus.http.port=8081 -Dquarkus.http.host=localhost</jvmArgs>
+          <jvmArgs>-Dquarkus.http.port=8081</jvmArgs>
         </configuration>
         <executions>
           <execution>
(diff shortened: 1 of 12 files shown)


ricardozanini (approved):

onboarding-example/hr/pom.xml
```diff
@@ -49,10 +49,20 @@
...
+            <quarkus.http.test-port>8081</quarkus.http.test-port>
+          </systemProperties>
+        </configuration>
       </plugin>
       <plugin>
         <groupId>io.quarkus</groupId>
         <artifactId>quarkus-maven-plugin</artifactId>
+        <configuration>
+          <debug>false</debug>
+          <jvmArgs>-Dquarkus.http.port=8081 -Dquarkus.http.host=localhost</jvmArgs>
```
cristianonicolai: is _quarkus.http.host_ actually needed? Default to 0.0.0.0 should be ok

onboarding-example/hr/pom.xml
```diff
@@ -49,10 +49,20 @@
     <plugins>
       <plugin>
         <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <systemProperties>
+            <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+            <quarkus.http.test-port>8081</quarkus.http.test-port>
```
radtriste: Why do we need to set the port on surefire plugin ?
the tests on the different projects are executed in parallel ?

onboarding-example/hr/pom.xml
jomarko: good point, fixed

onboarding-example/hr/pom.xml
jomarko: tests by default run 8081, but as we have three modules and one of them uses 8081 for real app it caused conflicts, this way each app uses same port for real app and tests so conflict shouldn't happen

jomarko: @radtriste please let me know my message above makes sense

radtriste: my question would be: why do we have a conflict during the tests ?

jomarko: ok, let me try to explain

before my changes
- I started module `hr` using `mvn clean package quarkus:dev`
- tests started on 8081, finished, dev mode remained active on 8080
- I started module `payroll` using `mvn clean package quarkus:dev`
- tests started on 8081, finished, dev mode failed because 8080 is used by `hr`

with my changes the conflict doesn't happen

Is my usage wrong? I mean I followed command for starting modules from readme

if that is the case please let e know I thikn we can then update readme files.

radtriste: ah now I get it. so you just want the test and the app to be on the same port.

Couldn't we just put directly the `quarkus.http.port` into `src/main/resources/application.properties` file from each project instead ?
we then would not need to set the port here or ? I think that file would be taken both by maven plugin when starting the app and during the test.

cristianonicolai: +1 to add to application.properties

jomarko: @radtriste @cristianonicolai thank you for application.properties advice. I like this way! so I applied. 

However still needed to distinguish `quarkus.http.port` and `quarkus.http.test-port`, without `quarkus.http.test-port` tests always tried to use 8081 regardless `quarkus.http.port` value.

cristianonicolai: @jomarko that's fine to also define the `test-port` but just bump with +1, as that helps running the tests without shutting down the app if running. I've done similar approach in Data Index, see: https://github.com/kiegroup/kogito-apps/blob/master/data-index/data-index-service/src/main/resources/application.properties#L11

jomarko: @cristianonicolai good point, applied

radtriste: sorry for being a PITA but could we put the `quarkus.http.test-port` property into `src/test/resources/application.properties` ?

jomarko: @radtriste sure

jomarko: changes are pushed

radtriste (approved):
I am good with it :)

onboarding-example/hr/pom.xml
jomarko: rebased

cristianonicolai (approved):

jomarko:
@evacchi hello, just friendly reminder, if you have a chance to review. Thank you.

evacchi (approved):

PR #194, opened 2020-04-08
Kogito Travel Agency support for Business Key

cristianonicolai:
- Auto generated travel code
- Added missing changes from previous demos: favicon and cors


kogito-travel-agency/travels/src/main/resources/META-INF/resources/index.html
```diff
@@ -107,11 +108,22 @@ <h1>Travels</h1>
...
 
     function create(travelRequest) {
-        $.post("/travels", travelRequest, function () {
+        var key = makeid();
+        $.post("/travels?businessKey=" + key, travelRequest, function () {
             load();
         }, "json");
     }
 
+    function makeid() {
```
ricardozanini, line 117: https://github.com/uuidjs/uuid

:stuck_out_tongue:

ricardozanini (approved):

kogito-travel-agency/travels/src/main/resources/META-INF/resources/index.html
cristianonicolai: I'm actually generating something like a reservation code ( 6 digits upper case ), like the one we get when doing a flight booking :) so not exactly the same thing as uuid

ricardozanini: I'm just annoying you!

AjayJagan (approved):
cool! <3

PR #167, opened 2020-03-30
KOGITO-1607 Rename bpmn samples to bpmn2

jomarko:
Unify business process file extensions.
For more details see https://issues.redhat.com/browse/KOGITO-1607

PR #202, opened 2020-04-09
[0.9.x] KOGITO-1607: unify bpmn2 vs bpmn file extension

jomarko:
Corollary of #167


radtriste (changes requested):
files are deleted ... is that on purpose ?
shouldn't they be migrated to bpmn2 ?

jomarko:
Yes it is on purpose. I guess during some conflict resolving we created copies of files by mistake. If you check process-quarkus-example and process-springboot-example after my change, they should contain just bpmn2 processes

I opened 4 PRs for 0.9.x, all are just cherry-picks, but this is "fix"

radtriste (approved):

ricardozanini:
@jomarko just checking, but during my renaming process, I saw much more `bpmn` files in the repo. Do we only have these?

jomarko:
@ricardozanini yes we still have as **bpmn** so **bpmn2** files in **kogito-examples**. My point was at least to do not mix in particular projects. In other words projects of **kogito-examples** now use either **bpmn2** or **bpmn**.

I would be for unification across whole **kogito-examples** however would need to be discussed with DOC team as it would affect them.

Even this PR was created as corollary as DOC review.

jstastny-cz (approved):

PR #208, opened 2020-04-09
KOGITO-1798: Disable example failing tests with persistence until def…

Sgitario:
…ects are fixed

JIRA Ticket: https://issues.redhat.com/browse/KOGITO-1798
Description: We have to ignore a couple of tests because some defects with persistance.
Also, we wanted to make the quarkus and spring boot versions behave the same using: same name files and using JUnit 5.
Finally, I added some configuration I needed for running Infinispan in Spring Boot (in Quarkus, this configuration was not necessary).


sutaakar:
@Sgitario IMHO it would be better to introduce a test category disabling the specified tests just when persistence is applied. Otherwise we lose test coverage.

Sgitario:
We cannot use the test category here since the same test case works for either persistence or not persistence.
In order to disable the tests, I made use of the surefire plugin to add a system property and then disable the concrete test methods using the @DisabledIfSystemProperty from JUnit 5.

sutaakar:
one side note, can you please use spaces in pom.xml rather than tabs, to keep content and formatting consistent?

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersProcessTest.java
```diff
@@ -0,0 +1,129 @@
+package org.kie.kogito.examples.demo;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
sutaakar, line 5: Can you please unify assetions?
You seems to use old and new JUnit at the same time.

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersProcessTest.java
```diff
@@ -0,0 +1,129 @@
...
+		assertEquals(1, orderItemProcesses.values().size());
+
+		ProcessInstance<?> childProcessInstance = orderItemProcesses.values().iterator().next();
+
+		List<WorkItem> workItems = childProcessInstance.workItems(policy);
+		assertEquals(1, workItems.size());
+
+		childProcessInstance.completeWorkItem(workItems.get(0).getId(), null, policy);
+
+		assertEquals(ProcessInstance.STATE_COMPLETED, childProcessInstance.status());
```
sutaakar, line 75: Why did you delete `assertEquals(ProcessInstance.STATE_COMPLETED, processInstance.status());`

Sgitario: It was failing and looking at the Quarkus version, this line was removed as part of https://issues.redhat.com/browse/KOGITO-173.

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersProcessTest.java
Sgitario: Done

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersProcessTest.java
sutaakar: hmm, ok
but why it doesn't fail in the test below? The line is there.

process-quarkus-example/pom.xml
```diff
@@ -129,6 +129,19 @@
...
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <systemPropertyVariables>
+                <tests.category>persistence</tests.category>
+              </systemPropertyVariables>
+            </configuration>
+	      </plugin>
```
sutaakar, line 142:
```suggestion
	    </plugin>
```

process-springboot-example/pom.xml
```diff
@@ -119,6 +119,9 @@
           <name>persistence</name>
         </property>
       </activation>
+      <properties>
+      	<tests.category>persistence</tests.category>
```
sutaakar:
```suggestion
        <tests.category>persistence</tests.category>
```

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersProcessTest.java
Sgitario: I didn't take a look, but I'm sure this depends on retriggering processes.

process-quarkus-example/pom.xml
Sgitario: Done

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersProcessTest.java
sutaakar: ok, can you please take a look then?

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersRestTest.java
```diff
@@ -0,0 +1,303 @@
...
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.annotation.DirtiesContext.ClassMode;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+import io.restassured.RestAssured;
+import io.restassured.http.ContentType;
+
+@ExtendWith(SpringExtension.class)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, classes = DemoApplication.class)
+@DirtiesContext(classMode = ClassMode.BEFORE_EACH_TEST_METHOD) // reset spring context after each test method
```
sutaakar: Wouldn't it have more sense to keep there AFTER_EACH_TEST_METHOD?

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersRestTest.java
```diff
@@ -0,0 +1,303 @@
...
+
+	@LocalServerPort
+	int port;
+
+	@BeforeEach
+	public void setUp() {
+		RestAssured.port = port;
+
+        // need it when running with persistence
+        orderProcess.instances().values().forEach(pi -> pi.abort());
```
sutaakar, line 51: IMHO it should be rather placed in AfterEach so the test cleans up after it has been executed.

Sgitario: In Quarkus version, it was being done before each test and it does not really matter if done before or after....

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersRestTest.java
Sgitario: changed .....

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersRestTest.java
sutaakar: Yes, though it may affect other tests when the cleanup is done in Before phase.

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersProcessTest.java
Sgitario: Raised: https://issues.redhat.com/browse/KOGITO-1848

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersRestTest.java
Sgitario: It's not going to be an issue since no more tests are using the order process, but will do.

sutaakar (approved):

process-springboot-example/src/main/resources/application.properties
```diff
@@ -2,4 +2,8 @@ spring.mvc.servlet.path=/docs
 
 resteasy.jaxrs.scan-packages=org.kie.kogito.**
 
-infinispan.remote.server-list=127.0.0.1:11222
\ No newline at end of file
+infinispan.remote.server-list=127.0.0.1:11222
```
radtriste: did you add the auth yourself ?

process-springboot-example/pom.xml
radtriste: any reason for double `tests.category` ?

radtriste (commented):
2 small comments below

process-springboot-example/src/main/resources/application.properties
Sgitario: Removed ...

process-springboot-example/pom.xml
Sgitario: fixed

radtriste (approved):

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersRestTest.java
cristianonicolai: I usually prefer to have the clean up done in the before method. That helps if you run let's say a separated service ( infinispan ? ) so the state is always the expected one when the test executes.

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersProcessTest.java
```diff
@@ -0,0 +1,129 @@
+package org.kie.kogito.examples.demo;
```
cristianonicolai, line 1: missing headers?

Sgitario: I don't see any header in the process-quarkus-example and process-springboot-example projects at least. Should we add them to all the sources? If so, I guess this should be done as part of another task. What do you think about?

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersRestTest.java
Sgitario: I agree with you. Usually, I do this stuff in the before methods as a setup configuration of my test suite. What do you think about @sutaakar ? Let me know whether reverting this or keeping it like this.

sutaakar: @Sgitario I don't have a strong preference in that, if you think it is better to do the cleanup before then feel free to do so.

Sgitario: In order to avoid to have an afterEach and a beforeEach both in the Spring Boot examples, I've changed it to be done in the beforeEach method as at the beginning.

process-springboot-example/src/test/java/org/kie/kogito/examples/demo/OrdersProcessTest.java
Sgitario: Raised a ticket for adding the headers: https://issues.redhat.com/browse/KOGITO-1866

cristianonicolai (approved):

PR #198, opened 2020-04-08
KOGITO-1644 Move docker tests to failsafe

jstastny-cz:
Keeping testcontainers, but moving docker dependent tests to integration phase using failsafe plugin.

PR #220, opened 2020-04-14
[0.9.x] KOGITO-1644 Move docker tests to failsafe

jstastny-cz:
* KOGITO-1644 Move docker tests to failsafe

cherry-pick of #198


process-usertasks-with-security-oidc-springboot/pom.xml
```diff
@@ -13,6 +13,11 @@
   <name>Kogito Examples :: Process Usertasks With Security OIDC Spring Boot</name>
   <description>Kogito usertasks orchestration with security enabled on REST api - Spring Boot</description>
 
+  <properties>
+    <container.image.keycloak>quay.io/keycloak/keycloak:${container.image.keycloak.version}</container.image.keycloak>
```
cristianonicolai, line 17: perhaps move this to root pom so it can be reused?

jstastny-cz: hm, that's what is in master atm .... I can do it master in a separate PR.

jstastny-cz: I mean in this PR it's not needed. If there are other examples to be added later, the PR adding the new ones will refactor this (but probably not on 0.9.x).

evacchi: if this is not strictly required for 0.9.x let's keep this for master only

evacchi (approved):

process-usertasks-with-security-oidc-springboot/pom.xml
jstastny-cz: Tracking the proposal as KOGITO-1867 and targeting 0.10.0

PR #221, opened 2020-04-14
KOGITO-1472 Random test failures on process-timer-quarkus example

tiagodolphine:
- update quarkus 1.3.1, that should have the fix
- disable jobs-service integration by default (to allow tests to be enabled)
- re-enable tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,7 +39,7 @@
     <surefire.forkCount>1</surefire.forkCount>
     <alphanetworkCompilerEnabled>false</alphanetworkCompilerEnabled>
 
-    <version.io.quarkus>1.3.0.Final</version.io.quarkus>
+    <version.io.quarkus>1.3.1.Final</version.io.quarkus>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -39,7 +39,7 @@
     <surefire.forkCount>1</surefire.forkCount>
     <alphanetworkCompilerEnabled>false</alphanetworkCompilerEnabled>
 
-    <version.io.quarkus>1.3.0.Final</version.io.quarkus>
+    <version.io.quarkus>1.3.1.Final</version.io.quarkus>
```
ricardozanini, line 42: Are we going to upgrade (or already have) the runtimes as well?

tiagodolphine: the runtime is already using 1.3.1, https://github.com/kiegroup/kogito-runtimes/blob/29b026040f247cab8a5918a7ce138674a29c5434/pom.xml#L113

cristianonicolai:
will retrigger the test a few times to check.

evacchi (approved):

ricardozanini (approved):
Should we cherry-pick for 0.9.x as well?

cristianonicolai (approved):

cristianonicolai:
> Should we cherry-pick for 0.9.x as well?

I guess we don't need to. upgrate to 1.3.1 is only done in master

PR #225, opened 2020-04-15
[KOGITO-1864] Adjust process-kafka-quickstart-springboot example for …

sutaakar:
…Kogito operator

Topics are created automatically.
Tested on OCP 4.3 and locally.

Signed-off-by: Karel Suta <ksuta@redhat.com>

### DIFF
--- a/process-kafka-quickstart-springboot/README.md
+++ b/process-kafka-quickstart-springboot/README.md
@@ -73,18 +73,6 @@ This quickstart requires an Apache Kafka to be available and by default expects 
 
 https://kafka.apache.org/quickstart
 
-In addition to that two topics are needed
-
-* travellers
-* processedtravellers
-
-```sh
-bin/kafka-topics.sh --create --bootstrap-server localhost:9092 --replication-factor 1 --partitions 1 --topic travellers
-bin/kafka-topics.sh --create --bootstrap-server localhost:9092 --replication-factor 1 --partitions 1 --topic processedtravellers
-```
-
-These topics are expected to be without key
-
 ## Build and run
 
 ### Prerequisites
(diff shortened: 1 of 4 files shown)


process-kafka-quickstart-springboot/README.md
```diff
@@ -73,18 +73,6 @@ This quickstart requires an Apache Kafka to be available and by default expects
 
 https://kafka.apache.org/quickstart
 
-In addition to that two topics are needed
```
radtriste, line 76: this part is no more accurate ?

sutaakar: No, topics are created automatically, user doesn't have to do anything.

jstastny-cz:
@sutaakar the example does not use Spring Boot Addon[1] we have, is that expected? 

[1] https://github.com/kiegroup/kogito-runtimes/tree/master/addons/events/kogito-events-spring-boot-addon

sutaakar:
@jstastny-cz That addon is not required as it collects process and task related events, they are usually processed by Data index.
Here we have just a simple example without any external integration with other components. IMHO the addon would have sense in case integration with Data index is implemented.

jstastny-cz:
@sutaakar well, if we deploy the example in OpenShift together with data-index and management-console. The processes won't be visible there now? I thought that was the main goal of the example to show exactly that integration ... otherwise we have IMO just a single example with events and that process-quarkus-example .... except of the travel agency app.

sutaakar:
Well, this process doesn't have any user task or persistence - you just send a JMS message and get different message back.
IMHO there won't be anything to be shown in Data index or Management console as it shows only existing processes and tasks (right?).

jstastny-cz:
Ah, ok, in that case I was wrong about the purpose of the example :-) thanks

radtriste (approved):

Sgitario (approved):

ricardozanini (approved):
Well done! Many thanks for this.

PR #227, opened 2020-04-15
Travel Agency - Add add endpoint for simulating a timeout error in th…

cristianonicolai:
…e process

### DIFF
--- a/kogito-travel-agency/travels/src/main/java/org/acme/travels/service/FlightBookingResource.java
+++ b/kogito-travel-agency/travels/src/main/java/org/acme/travels/service/FlightBookingResource.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates. 
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.acme.travels.service;
+
+import javax.inject.Inject;
+import javax.ws.rs.GET;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+@Path("/flightservice")
+public class FlightBookingResource {
+
+    @Inject
+    FlightBookingService service;
+
+    @Path("/shouldtimeout")
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    public Boolean getShouldTimeout() {
+        return service.getShouldTimeout();
+    }
+
+    @Path("/shouldtimeout")
+    @PUT
+    @Produces(MediaType.APPLICATION_JSON)
+    public Boolean toggleTimeout() {
+        service.setShouldTimeout(!service.getShouldTimeout());
+        return getShouldTimeout();
+    }
+}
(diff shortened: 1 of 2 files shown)


kogito-travel-agency/travels/src/main/java/org/acme/travels/service/FlightBookingService.java
```diff
@@ -1,14 +1,38 @@
...
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
 
 import org.acme.travels.Flight;
 import org.acme.travels.Trip;
 
 @ApplicationScoped
+@Path("/flightservice")
```
radtriste: Could you create another class that handle that and get the `FlightBookingService` bean ?

kogito-travel-agency/travels/src/main/java/org/acme/travels/service/FlightBookingService.java
```diff
@@ -1,14 +1,38 @@
...
 public class FlightBookingService {
 
-	public Flight bookFlight(Trip trip) {
-		return new Flight("MX555", trip.getBegin(), trip.getEnd());
-	}
+    private Boolean timeout = Boolean.FALSE;
```
radtriste: `shouldTimeout` ?

kogito-travel-agency/travels/src/main/java/org/acme/travels/service/FlightBookingService.java
```diff
@@ -1,14 +1,38 @@
...
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    public Boolean getTimeout() {
+        return timeout;
+    }
+
+    @POST
+    @Produces(MediaType.APPLICATION_JSON)
+    public boolean setTimeout() {
```
radtriste: `toggleTimeout()` ?

radtriste (commented):
see my comments below

tiagodolphine (commented):
Just a suggestion, even this is a simple endpoint, what about creating a separated class to expose the endpoints something like `FlightBookingResource` and this class injects the `FlightBookingService` where set/get the timeout flag.

kogito-travel-agency/travels/src/main/java/org/acme/travels/service/FlightBookingService.java
tiagodolphine: +1

cristianonicolai:
@tiagodolphine @radtriste comments addressed.

kogito-travel-agency/travels/src/main/java/org/acme/travels/service/FlightBookingResource.java
```diff
@@ -0,0 +1,29 @@
...
+    @Inject
+    FlightBookingService service;
+    
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    public Boolean getShouldTimeout() {
+        return service.getShouldTimeout();
+    }
+
+    @POST
```
radtriste: I would be this as a `PUT` method as it is more an update of an existing thing.

kogito-travel-agency/travels/src/main/java/org/acme/travels/service/FlightBookingResource.java
```diff
@@ -0,0 +1,29 @@
...
+import javax.ws.rs.core.MediaType;
+
+@Path("/flightservice")
+public class FlightBookingResource {
+
+    @Inject
+    FlightBookingService service;
+    
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
```
radtriste, line 34: Sorry for being a PITA but still not a big fan to get/update a boolean value on `/flightservice` path.
Could we have something like `/flightservice/timeout` or `/flightservice/shouldtimeout` ?

or if want to keep `/flightservice`, then we should get/update a configuration structure. but it is more complex.

tiagodolphine: I agree, adding to `/flightservice/shouldtimeout` is even more flexible in case we need to add more actions to the `flightservice` other than the timeout.

ricardozanini (approved):
Once @radtriste's comments were addressed, I'm good with it. :)

radtriste (approved):

cristianonicolai:
@tiagodolphine @radtriste updated :)

tiagodolphine (approved):

PR #229, opened 2020-04-16
KOGITO-1866 Add Red Hat headers to all sources in Kogito Examples

bxf12315:


bxf12315:
@cristianonicolai please check it.

decisiontable-quarkus-example/pom.xml
```diff
@@ -1,4 +1,21 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!--
```
cristianonicolai: I dont think we need to add it to xml, only Java classes.

drools-quarkus-example/kubernetes/service.yml
```diff
@@ -1,3 +1,19 @@
+#
```
cristianonicolai: @ricardozanini would these files ( yml ) need headers?

jbpm-optaplanner-springboot-example/mvnw.cmd
```diff
@@ -1,21 +1,18 @@
-@REM ----------------------------------------------------------------------------
-@REM Licensed to the Apache Software Foundation (ASF) under one
```
cristianonicolai: not sure we should change these, they are autogenerated and already included headers.
@ricardozanini @evacchi wdyt ?

cristianonicolai (commented):
I think changes should be limited to Java classes but I'm not an expert in the subject.
@krisv @evacchi @ricardozanini @tiagodolphine  wdyt ?

jbpm-optaplanner-springboot-example/mvnw.cmd
radtriste: those mvnw are not our code. this was taken from outside. we should not change it IMO

decisiontable-quarkus-example/pom.xml
radtriste: yeah ... what we really produce is the Java code. not sure xml files need the header ...

drools-quarkus-example/kubernetes/service.yml
radtriste: how is it done on other kiegroup repo ?

jbpm-optaplanner-springboot-example/mvnw.cmd
bxf12315: I use license-maven-plugin so can exclude xml and cmd files

bxf12315:
@cristianonicolai updated, Please review again.

jbpm-optaplanner-springboot-example/mvnw.cmd
cristianonicolai: @bxf12315 perhaps that could be added so the project fails to compile if the header is missing on a Java class?

bxf12315: ok I will take a try

radtriste: Or we have a way to add it automatically on compilation ?

drools-quarkus-example/kubernetes/service.yml
ricardozanini: Now they don't need it, also if IIRC, `oc client` doesn't like those comments in the yaml files.

jbpm-optaplanner-springboot-example/mvnw.cmd
ricardozanini: Generated code taken from outside should not have license headers.

decisiontable-quarkus-example/src/main/java/org/kie/kogito/queries/Applicant.java
```diff
@@ -1,18 +1,18 @@
-/*
- * Copyright 2019 Red Hat, Inc. and/or its affiliates.
```
ricardozanini, line 2: we should not update the year

ricardozanini (changes requested):
Regarding the year update, it's a nitpick... but you should open this to the `master` branch instead. Those examples were renamed and some of them does not exist any more.

Do we have a way to automatically add those headers? For example, in the operator every time you run the unit tests, we have a script that adds the header if it's not there.

cristianonicolai:
> Regarding the year update, it's a nitpick... but you should open this to the `master` branch instead. Those examples were renamed and some of them does not exist any more.
> 
> Do we have a way to automatically add those headers? For example, in the operator every time you run the unit tests, we have a script that adds the header if it's not there.

ouch, good catch, definitely needs to go into _master_ not _stable_ branch.

decisiontable-quarkus-example/src/main/java/org/kie/kogito/queries/Applicant.java
bxf12315: yes, The license automatically by license-maven-plugin, you just need run mvn license:format[1].
but I not sure we can skip year's update. it's a little bit hard to do.
1 https://github.com/mycila/license-maven-plugin

ricardozanini: Don't mind. But I fear you will have to do it all again since this branch it's not supposed to be PR'ed. See my comment below. :(

bxf12315: ok

radtriste:
@bxf12315 You need to recreate a PR against `master` branch
You should not have a PR against branch `stable`.

I close this one and let you recreate one against `master` branch

PR #244, opened 2020-04-22
[KOGITO-1954] Add examples of DMNRuntimeEventListener injection

kostola:


danielezonca (approved):

tarilabs (approved):

r00ta:
What about adding some tests so that this example works as integration test as well?

tarilabs:
@r00ta +1 for adding tests, but I would avoid doing it in these examples. I believe this example should be kept as simple as possible. I believe is best to have an integration test in `kogito-runtimes.git`

kostola:
> +1 for adding tests, but I would avoid doing it in these examples

I expanded the tests as suggested by @r00ta before reading this comment by @tarilabs :man_facepalming: 

Anyway it's not a big change and it assures that listeners are truly injected both in Quarkus and Spring Boot. Also the end user could see how to access the injected listeners in `Config` object by looking at the tests.

@tarilabs and @danielezonca please take a look. If you don't like the change I can remove the last commit and we can merge the original approved version.

danielezonca (approved):
LGTM :)

tarilabs:
@kostola trying to find a compromise, because imho is now difficult to see the simplicity in the test of DMN example, where people are coming to see how to use DMN, not the listener... my suggestion if you want to keep the test in this example (which is fine) is to move the test into a separate test method, FOLLOWING the currently existing one which shall remain invaried.
In other words people will access the test and will see 2 unit test `@Test`-annotated method.
The first one, shall be the one on master today.
The second one, exercise (also) the listener.
wdyt?

tarilabs:
Or even better you can copy-paste the class into another JUnit class
up to you really.

tarilabs (commented):
commented suggested test re-organizations

evacchi (approved):
good for me when/if all suggestions by @tarilabs are addressed

kostola:
@evacchi @danielezonca I moved the tests related to listener injection in separate classes as suggested by @tarilabs . I think this can be merged now.

danielezonca:
@MarianMacik @radtriste 
Can you please have a look?

tarilabs (approved):

radtriste:
I am a bit concerned about the complexity that it brings to the example, thinking about the user coming the first time to this example. Even if it is written `optional` into the README, I think we should not put too much thing into a simple example.

Wouldn't it be better to create another example with this listener ? 
As it is done for process-usertasks-*, we could call it dmn-custom-logic-*-example.
So we keep simplicity into the dmn-quarkus-example/dmn-springboot-example and we have more complex ones with maybe later some other added functionalities in it.

I know this does bring copy/paste of code but what we want is for the user to understand quickly what happens here.

dmn-springboot-example/README.md
```diff
@@ -26,6 +26,17 @@ mvn clean package
...
+
+Kogito allows you to inject custom instances of `DMNRuntimeEventListener` if you need to attach custom logic to every DMN model evaluation.
+
+There are two ways to do this:
+* Create one or more standard listener classes and annotate them with `Component` (the quickest way to inject a single listener). Demonstrated in `ExampleDMNRuntimeEventListener` class.
+* Create one or more instances of `DecisionEventListenerConfig` (returning a list of listeners each) and annotate them with `Component`. Demonstrated in `ExampleDecisionEventListenerConfig` class.
+
+All the listeners instantiated with both methods will be injected during the application startup phase.
+
+**Important:** everything that this section describes is *OPTIONAL*: if you don't need listeners, just ignore it.
```
MarianMacik: I would unify this note across both examples.

dmn-springboot-example/src/main/resources/Bikes.dmn
```diff
@@ -0,0 +1,306 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<dmn:definitions xmlns:dmn="http://www.omg.org/spec/DMN/20180521/MODEL/" id="_3ECBF416-244D-4359-8655-C3D7B1B40EE3"
+                 name="Bikes" typeLanguage="http://www.omg.org/spec/DMN/20180521/FEEL/"
```
MarianMacik: Is this file needed for traffic violation?

dmn-springboot-example/src/main/resources/Loan Eligibility.dmn
```diff
@@ -0,0 +1,172 @@
+<dmn:definitions xmlns:dmn="http://www.omg.org/spec/DMN/20180521/MODEL/" xmlns="https://github.com/kiegroup/kogito-examples/dmn-quarkus-listener-example" xmlns:di="http://www.omg.org/spec/DMN/20180521/DI/" xmlns:kie="http://www.drools.org/kie/dmn/1.2" xmlns:dmndi="http://www.omg.org/spec/DMN/20180521/DMNDI/" xmlns:dc="http://www.omg.org/spec/DMN/20180521/DC/" xmlns:feel="http://www.omg.org/spec/DMN/20180521/FEEL/" id="_DC5AB7D9-7D61-40F7-94C1-2E064A053AC0" name="Loan Eligibility" typeLanguage="http://www.omg.org/spec/DMN/20180521/FEEL/" namespace="https://github.com/kiegroup/kogito-examples/dmn-quarkus-listener-example">
+  <dmn:extensionElements/>
+  <dmn:itemDefinition id="_AD041308-CE7E-4359-97DA-E03C520A7A70" name="tClient" isCollection="false">
+    <dmn:itemComponent id="_91AB76A9-CDCC-463E-B404-C530AB0AA6B5" name="age" isCollection="false">
```
MarianMacik: Same here, is this needed for traffic viloation?

MarianMacik (changes requested):
@radtriste I believe that the added DMN definitions are really not needed. Then it is just a few classes added + some tests, so the change itself is not that big.

Still, we should consider if this should be a part of a basic DMN example.

dmn-springboot-example/src/main/resources/Bikes.dmn
kostola: Damn...this is a custom model I added for personal tests.
I pushed it by mistake and will remove it immediately. Sorry! :man_facepalming:

dmn-springboot-example/src/main/resources/Loan Eligibility.dmn
kostola: Same as `Bikes.dmn`...will remove it!

dmn-springboot-example/README.md
kostola: Ok, shall I use this exact sentence also in Quarkus README then?

kostola:
@radtriste @MarianMacik @evacchi @danielezonca @tarilabs It's easy to move all this logic into new examples instead of extending the existing ones as I did now. Please let's reach a final agreement so this can be merged.

This PR was initially supposed to be very small, which justified the change to the existing examples. Now, with the addition of tests, it would be better imho to create new examples (like `dmn-quarkus-listener-example` and `dmn-springboot-listener-example`) to avoid generating confusion in people approaching Kogito for the first time.

Wdyt?

dmn-springboot-example/src/main/resources/Bikes.dmn
kostola: Removed

dmn-springboot-example/src/main/resources/Loan Eligibility.dmn
kostola: Removed

dmn-springboot-example/README.md
MarianMacik: I would use the one from Quarkus also here :) But really it doesn't matter as long as it is the same.

MarianMacik:
OK, then I think you can go for separate examples.

dmn-springboot-example/README.md
kostola: @MarianMacik I didn't use the same sentence as Quarkus because it says _"you can have a Quarkus application that runs DMN models on Kogito without writing a single line of Java code"_ but in the Spring Boot application there's one Java class (`KogitoSpringbootApplication`) :sweat_smile:

kostola:
@MarianMacik moved to separate examples :+1:

dmn-springboot-example/README.md
kostola: Now that we're using separate examples I wrote a sentence regarding the optionality of listeners directly in the _Description_ section of each `README.md` file (the same for both as you requested).

radtriste:
@kostola can you check the failures of the check ?
Seems your projects are failing

```
Failed to execute goal com.mycila:license-maven-plugin:3.0:check (default) on project dmn-listener-quarkus: Some files do not have the expected license header -> [Help 1]

[ERROR] Failed to execute goal com.mycila:license-maven-plugin:3.0:check (default) on project dmn-listener-springboot: Some files do not have the expected license header -> [Help 1]
```

kostola:
@radtriste @MarianMacik everything seems fine now (the license checker is quite harsh btw :smirk:).
Idk how to mark the change request as done but all the comments are resolved.

radtriste (approved):

MarianMacik (approved):
Looks good now.

PR #245, opened 2020-04-22
Add Scenario Simulation to dmn-springboot-example

dupliaka:
It is possible to work with Scesim files inside VSCode.
Test Scenario Simulation works for DMN and available to use
if user includes Kogito ScenarioJunitActivator and
adds related scesim file to the project + add kogito scenario simulation dependency.

I made this example to look exactly as it is implemented in Business central to make it familiar for users.


dupliaka:
Closed, 
cloned to merge into master :
https://github.com/kiegroup/kogito-examples/pull/246

PR #258, opened 2020-05-15
[Kogito-2195] Kogito examples with ansible automation

desmax74:
https://issues.redhat.com/browse/KOGITO-2195
@sterobin @danielezonca


sterobin:
@danielezonca , @krisv , @ricardozanini , @radtriste , who should we add as dev reviewers and QE reviewers (if applicable) for this to get it worked in and approved for the examples? Max and I apparently don't have permissions to add reviewers officially.

kogito-ansible/README.md
```diff
@@ -0,0 +1,57 @@
+# Openshift (CRC) Kogito ansible automation
+
+Ansible scripts to automate creation of CRC cluster and the deploy of one of the Kogito examples with just one command line.
+
+CRC 1.10.0, and Ansible must be installed.
+No previous CRC setup (no /home/{user}/.crc folder), otherwise the create script will fail, delete .crc if you want run more than once the create playbook
```
jiripetrlik: Please make this optional. User may want to use his already CRC deployed cluster. It is probably not good to force user to remove all his previous CRC setup.

kogito-ansible/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Create an account on cloud.redhat.com
+https://cloud.redhat.com/openshift/install and download or copy your Pull secret from the the laptop installation https://cloud.redhat.com/openshift/install/crc/installer-provisioned
+
+
+### Install CRC
+(If you haven't already installed)
+
+Pre requisite on Debian/Ubuntu:
```
jiripetrlik: Please provide version of Ubuntu and Debian on which you have tested these scripts.

kogito-ansible/README.md
```diff
@@ -0,0 +1,57 @@
...
+change the app_name (my-kafka-project) in the file if you want different name
+```sh
+ansible-playbook ./playbook_crc.yaml
+```
+
+Configure etc/hosts (default is kafka_cluster_name: "my-cluster-kafka" and app_namespace: "my-kafka-project")
+```sh
+sudo ansible-playbook ./playbook_etc_hosts.yaml
+```
+
```
jiripetrlik: Please provide setup also for Fedora.

kogito-ansible/playbook_libs.yaml
```diff
@@ -0,0 +1,11 @@
+---
+- name: Install libs on Debian/Ubuntu
+  hosts: localhost
+  gather_facts: true
+
+  tasks:
+
+    - name: install
+      shell: "sudo apt install qemu-kvm libvirt-daemon libvirt-daemon-system network-manager -y"
```
jiripetrlik: Please use "apt" module instead of shell. https://docs.ansible.com/ansible/latest/modules/apt_module.html#apt-module

kogito-ansible/playbook_libs.yaml
```diff
@@ -0,0 +1,11 @@
...
+      become: yes
+      when: ansible_os_family == 'Debian'
```
jiripetrlik: Please check that os family is supported using assert module instead of silently skipping installation of libvirt libraries.
https://docs.ansible.com/ansible/latest/modules/assert_module.html#assert-module

kogito-ansible/playbook_libs.yaml
```diff
@@ -0,0 +1,11 @@
+---
+- name: Install libs on Debian/Ubuntu
+  hosts: localhost
+  gather_facts: true
```
jiripetrlik: What is the reason for gather_facts here? As far as I know Ansible will gather facts automatically at start of the playbook.

kogito-ansible/playbook_create.yaml
```diff
@@ -0,0 +1,131 @@
...
+        private: "no"
+      - name: "PullSecret"
+        prompt: "Insert your Pull secret"
+
+    tasks:
+      - debug: msg="Project Name {{ nameSpace }}"
+
+      - name: Copy Kogito cli
+        get_url:
+          url: "https://github.com/kiegroup/kogito-cloud-operator/releases/download/{{ kogito_tag }}/kogito-cli-{{ kogito_tag }}-linux-amd64.tar.gz"
```
jiripetrlik: Can you please specify variable for this with default value to make it configurable.

kogito-ansible/playbook_create.yaml
```diff
@@ -0,0 +1,131 @@
...
+          dest: "/tmp/kogito-{{ kogito_tag }}-linux-amd64.tar.gz"
```
jiripetrlik: Again please create variable for this temporary destination with default value. Or you can also consider to use "tempfile" module. https://docs.ansible.com/ansible/latest/modules/tempfile_module.html#tempfile-module

kogito-ansible/playbook_etc_hosts.yaml
```diff
@@ -0,0 +1,28 @@
...
+  hosts: localhost
+
+  vars:
+    kafka_cluster_name: "my-cluster-kafka"
+    app_namespace: "my-kogito-project"
+    ip_on_etc_hosts: 192.168.130.11
+
+  tasks:
+
+  - name: Check api in hosts
```
jiripetrlik: Isn't DNS reconfigured by CRC itself.

kogito-ansible/playbook_create.yaml
```diff
@@ -0,0 +1,131 @@
+---
+  - name: Start CRC 1.10.0 and deploy Kogito Operator and CLI
+    hosts: localhost
+    gather_facts: true
+
+    vars:
+      kogito_tag: "0.10.0"
+      ram: "16384" #MB
+      libvirt_version: "4.4.3"
+      ip_on_etc_hosts: 192.168.130.11
```
jiripetrlik: Please do not use this value which can probably change on different machine etc. It would be much better to use "virt" or "virt_net" modules to obtain info about IP address.

- https://docs.ansible.com/ansible/latest/modules/virt_module.html#virt-module
- https://docs.ansible.com/ansible/latest/modules/virt_net_module.html#virt-net-module

kogito-ansible/playbook_create.yaml
```diff
@@ -0,0 +1,131 @@
+---
+  - name: Start CRC 1.10.0 and deploy Kogito Operator and CLI
+    hosts: localhost
+    gather_facts: true
```
jiripetrlik: Do we really need to explicitly run this?

kogito-ansible/playbook_crc.yaml
```diff
@@ -0,0 +1,55 @@
+---
+- name: Install CRC
+  hosts: localhost
+  gather_facts: true
```
jiripetrlik: Isn't this done by default?

jiripetrlik (changes requested):
@desmax74 
Hello! Thank you for your PR. I've put the first review here. Unfortunately I was not able to test it because the setup is for Ubuntu/Debian. Can you please add setup also for some RedHat like systems such as Fedora?

jiripetrlik:
@sterobin Hello I did the first QE review for this.

desmax74:
@jiripetrlik These scripts are developed and tested on Fedora 31, only the playbook_scripts is for ubuntu if is the system used.

jiripetrlik:
@desmax74 It seems to me that "playbook_libs.yaml" is developed for Ubuntu isn't it? And as libvirts setup is necessary for whole setup we probably need alternative for Fedora. Do you agree?

desmax74:
@jiripetrlik yep, only playbook_libs.yaml is for ubuntu, if isn't clear the description "Pre requisite on Debian/Ubuntu:
Install libvirt libs on Debian/Ubuntu only:" " I could change.
Libvirt was already present on Fedora iirc.

kogito-ansible/playbook_crc.yaml
desmax74: No, is disabled by default to save time, but to read the current user infos is needed

kogito-ansible/playbook_create.yaml
desmax74: No, was just to describe the version used, we could use a placeholder and print the version used instead to updated this on every update

kogito-ansible/playbook_create.yaml
desmax74: Ok

kogito-ansible/playbook_libs.yaml
desmax74: Ok

kogito-ansible/playbook_libs.yaml
desmax74: Ok

kogito-ansible/playbook_etc_hosts.yaml
desmax74: I have to check with the latest version, because with old version of crc without this manual addition on /etc/hosts the cluster was unreachable

kogito-ansible/playbook_create.yaml
desmax74: Ok

kogito-ansible/playbook_create.yaml
desmax74: Ok

kogito-ansible/README.md
desmax74: Ok

kogito-ansible/README.md
desmax74: Ok

kogito-ansible/README.md
desmax74: I'll create two sections in the doc since is all for fedora and only the lib virt installation is the first step on ubuntu

jiripetrlik:
@desmax74 
I would suggest to install "dnf install @virtualization" to be sure. Please use proper package module of Ansible to do that. Or check if libvirtd is running and in case that it does not install virtualization packages. To be honest don't know which packages are installed by default. But it may probably differ on Fedora type, such as desktop or server. Or it may be different for Fedora spins.

See documentation for Fedora virtualization packages: https://docs.fedoraproject.org/en-US/quick-docs/getting-started-with-virtualization/

kogito-ansible/README.md
```diff
@@ -0,0 +1,90 @@
+# Openshift (CRC) Kogito ansible automation
+
+Tested on Fedora 31/32 and Ubuntu 19.10
+
+Ansible scripts to automate creation of CRC cluster and the deploy of one of the Kogito examples with just one command line.
+
+Ansible must be installed.
```
danielezonca: Can you please add a link to https://docs.ansible.com/ansible/latest/installation_guide/index.html ?

kogito-ansible/playbook_crc.yaml
```diff
@@ -0,0 +1,56 @@
+---
+- name: Install CRC
+  hosts: localhost
+
+  vars:
+    crc_version: "1.10.0"
+    crc_url_file: "https://mirror.openshift.com/pub/openshift-v4/clients/crc/{{crc_version}}/crc-linux-amd64.tar.xz"
+    crc_tmp_dest: "/tmp/crc-linux-{{crc_version}}-amd64.tar.xz"
+    crc_package_name: "crc-linux-{{crc_version}}-amd64"
```
danielezonca: What about use `latest` path to have CRC version automatically updated ( https://mirror.openshift.com/pub/openshift-v4/clients/crc/latest/ )?
As alternative it could be nice to let the user prompts the version

kogito-ansible/playbook_create.yaml
```diff
@@ -0,0 +1,134 @@
+---
+  - name: Start CRC and deploy Kogito Operator and CLI
+    hosts: localhost
+
+    vars:
+      kogito_tag: "0.10.0"
```
danielezonca: Is it possible to get latest version from the project or update this value automatically? (a maven variable?)

kogito-ansible/playbook_create.yaml
```diff
@@ -0,0 +1,134 @@
...
+      ram: "16384" #MB
+      libvirt_version: "4.4.3"
```
danielezonca: Why is this value hardcoded?

README.md
```diff
@@ -108,3 +108,8 @@ shows timers (intermediate and boundary) that allows to introduce delays in proc
 Every example has a directory named `operator` including the YAML files to deploy it using the Kogito Operator in an OpenShift cluster.
 Please refer to the [Kogito Documentation](https://docs.jboss.org/kogito/release/latest/html_single/#chap_kogito-deploying-on-openshift)
 of how to install the operator to your environment in order to try it there.
+
+## Deploying the Kogito example applications on OpenShift CRC using Ansible playbooks
```
danielezonca: I think it could be useful to add `locally` in the title: I don't think everyone know what CRC is.
You can change it as 
`Deploying the Kogito example applications locally with OpenShift CRC using Ansible playbooks`

kogito-ansible/README.md
danielezonca: Can you please refactor this paragraph as `Prerequisite`?
Can you please add to this section also system requirement needs in terms of RAM and disk?

kogito-ansible/playbook_deploy.yaml
```diff
@@ -0,0 +1,55 @@
+---
+  - name: Start CRC 1.10.0 and deploy kogito examples
+    hosts: localhost
+
+    vars:
+      libvirt_version: "4.4.3"
```
danielezonca: Same comment about hardcoded version.

kogito-ansible/playbook_create.yaml
```diff
@@ -0,0 +1,134 @@
...
+          content: "{{ PullSecret }}"
+          dest: "~/.crc/crc-pull-secret"
+
+      - name: Copy Kogito Examples pom
+        copy:
+          src: "./../pom.xml"
+          dest: "/tmp/examples_pom.xml"
+
+      - name: "Get Kogito Examples "
+        shell: xmllint --xpath '/*[local-name()="project"]/*[local-name()="modules"]' /tmp/examples_pom.xml > ~/.crc/quarkus_examples.txt
```
danielezonca: Why `quarkus`? Is this script limited to quarkus example?
If yes why?
If not I would prefer to name it `kogito-examples.txt`

kogito-ansible/playbook_deploy.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+    vars_prompt:
+      - name: "nameSpace"
+        prompt: "Insert your Project Name/ NameSpace"
+        default: "my-kogito-project"
+        private: "no"
+
+      - name: "example"
+        prompt: "Choose the example to deploy"
+        default: "\n{{ examples }}"
```
danielezonca, line 23: Can you please clarify what does it mean?
By default will this deploy `all` kogito examples in the same namespace?

kogito-ansible/README.md
```diff
@@ -0,0 +1,90 @@
...
+To deploy one of the kogito examples, enter the name of one of the example,
+at the end the browser show you the login page, the output of the deploy with the instructions to see
+the logs from cli will be on the ansible output.
+```sh
+ansible-playbook ./playbook_deploy.yaml
+```
+
+
+
+## Ubuntu
```
danielezonca: Is this section actually different from the `Fedora` one?
They are really similar (the same?) from what I can see. Is it possible to create a single `Fedora/Ubuntu` section and just highlight different steps (if any).
This can reduce maintenance cost of this doc :)

danielezonca (commented):
Thanks for the PR :)
Some comments

kogito-ansible/playbook_libs.yaml
```diff
@@ -0,0 +1,25 @@
...
+        pkg:
+          - qemu-kvm
+          - libvirt-daemon
+          - libvirt-daemon-system
+          - network-manager
+      become: yes
+      when: ansible_os_family == 'Debian'
+
+    - name: Install Virtualization on Fedora
+      shell: "sudo dnf install @virtualization -y"
```
jiripetrlik: Please use Ansible DNF module instead https://docs.ansible.com/ansible/latest/modules/dnf_module.html#dnf-module

Something like this:
    - name: Install Virtualization on Fedora
      dnf:
        name: "@virtualization"
      become: yes
      when: ansible_os_family == 'RedHat'

kogito-ansible/README.md
desmax74: ok

README.md
desmax74: ok

kogito-ansible/README.md
desmax74: ok

kogito-ansible/playbook_create.yaml
desmax74: Because is used in the path and you discover the version only when the crc is downloaded and unpacked, or readed from the lates json info

kogito-ansible/playbook_deploy.yaml
desmax74: Because is used in the path and you discover the version only when the crc is downloaded and unpacked, or readed from the lates json info

kogito-ansible/playbook_create.yaml
desmax74: ok, I'll rename kogito-examples

kogito-ansible/playbook_deploy.yaml
desmax74: It deploy only the choosed example, default is only the message to display

kogito-ansible/README.md
desmax74: Previously only one section plus the ubuntu lib virt installation was present but at the reviewer seems a document for ubuntu deploy and asked a fedora version, for this reason you see  explicit configurations for both systems

kogito-ansible/playbook_libs.yaml
desmax74: ok

kogito-ansible/README.md
danielezonca: I agree to have the instructions for `Ubuntu` and `Fedora` too but if they are the same just mention that the instructions are tested on both configuration instead of duplicate them :)

kogito-ansible/playbook_create.yaml
desmax74: I've writed a version to use the latest kogito tag, but the 0.10.1 doesn't work because the cli and the operator aren't aligned  (kogito 0.10.1 - kogito-cloud-operator 0.10.0)

danielezonca: mmm I expect this could happen in the future too so is it possible to add a `kogito-operator` variable to support also this use case?

jiripetrlik:
@desmax74 
I've found two possible bugs when installing on fresh Ubuntu 20.04.

1. In playbook "playbook_crc", "crc libvirt group" task. Crc can not be found because folder ".local/bin" did not exist before running Ansible script, To workaround this it is necessary to reinitialize terminal and run command manually "crc config set skip-check-user-in-libvirt-group true"

2. In "playbook_create" in "Start CRC with 13312MB" task. I see the following error: "Failed to query DNS from host: lookup api.crc.testing: no such host"

desmax74:
@jiripetrlik 
1) Why not just simply create the folder as a pre-task ?

2) I could re-apply https://github.com/kiegroup/kogito-examples/pull/258/commits/654af19b9bd72e3ca642579ef3ac0e1786c13239#diff-5ab00d684a07bbdf26507b898ba2a2ff

jiripetrlik:
@desmax74 
ad 1, I've moved "Create .local/bin/" task to pre_tasks but without any effect. It is still necessary to login and logout to have crc command on PATH.
ad 2, I've tried to run "kogito-ansible/playbook_etc_hosts.yaml" before playbook_create but it failed with the following error "Failed to query DNS from host: lookup foo.apps-crc.testing: no such host". I think we need to find more general solution than adding lines to hosts file for each domain. Is CRC supported on Debian/Ubuntu?

desmax74:
In the Ubuntu 20.04 bashrc isn't present the `PATH="$HOME/.local/bin` 
I'll change the script to change the bashrc to re add like the 19.10

kogito-ansible/playbook_create.yaml
```diff
@@ -0,0 +1,169 @@
...
+    - name: "PullSecret"
+      prompt: "Insert your Pull secret"
+    - name: "ram"
+      default: "13312"
+      prompt: "Ram used in #MB"
+      private: "no"
+    - name: "kogitoTag"
+      default: "0.10.0"
+      prompt: "Kogito CLI and Operator tag"
+      private: "no"
```
danielezonca, line 22: Can you split in two different values? You can propose by default the same value user specified in the first step. Something like (I don't know Ansible syntax so it could be wrong)
```yaml
    - name: "kogitoTag"
      default: "0.10.0"
      prompt: "Kogito Version"
      private: "no"
    - name: "kogitoOperator"
      default: {{ kogitoTag }}
      prompt: "Kogito Operator version"
      private: "no"
```
Wdyt?

desmax74: Is up to kogito the installation of the operator and it didn't pass values like version, we ask the kogito version just to decide which version to download https://github.com/kiegroup/kogito-examples/pull/258/files#diff-a5f485f8a086810799bb19ca584ec530R150

danielezonca (approved):
Approving :+1: 
@desmax74 can you please rebase from master to make the build green?

desmax74:
@jiripetrlik dropped ubuntu until we found a workaround, please review

jiripetrlik:
@desmax74 Hello currently I'm getting following error on Fedora 32 : ""changed": true, "cmd": "crc setup", "delta": "0:03:00.977265", "end": "2020-06-16 10:31:41.176060", "msg": "non-zero return code", "rc": 1, "start": "2020-06-16 10:28:40.198795", "stderr": "level=fatal msg=\"Failed to start libvirt 'crc' network\"", "stderr_lines"...

jiripetrlik:
@desmax74 It also seems that there are some commits unrelated to PR and PR check failed. Can you please rebase properly?

domhanak:
@krisv @mdproctor @ricardozanini do you have any information about this?

ricardozanini:
I'd close it.

domhanak:
Closing as stale

PR #262, opened 2020-05-27
Test OptaPlanner Spring Boot example via REST API

rsynek:
- turns the existing test into an integration test that uses a REST API
- replaces hardcoded waiting for optaplanner with polling


Christopher-Chianelli (approved):

rsynek:
@Christopher-Chianelli if you're ok with that, I would make the same changes to the Quarkus version of the example.

PR #278, opened 2020-06-17
KOGITO-1867: Move docker image property definitions to parent pom

Sgitario:
JIRA: https://issues.redhat.com/browse/KOGITO-1867
Description:
- Remove the container keycloak configuration from examples and move them to the parent module.
- Normalize the Keycloak container as done in the Infinispan container.


sutaakar:
Does it mean that all examples will share same Keycloak realm?

Sgitario:
> Does it mean that all examples will share same Keycloak realm?

For integration tests, they will. However, it's possible to overwrite it in a concrete test.

cristianonicolai (approved):

sutaakar (approved):

radtriste (approved):

examples-test-utils/src/main/java/org/kie/kogito/testcontainers/KeycloakContainer.java
```diff
@@ -0,0 +1,51 @@
...
+        withEnv("KEYCLOAK_PASSWORD", "admin");
+        withEnv("KEYCLOAK_IMPORT", REALM_FILE);
+        withClasspathResourceMapping("testcontainers/keycloak/kogito-realm.json", REALM_FILE, BindMode.READ_ONLY);
+        withLogConsumer(new Slf4jLogConsumer(LOGGER));
+        waitingFor(Wait.forHttp("/auth").withStartupTimeout(Duration.ofMinutes(5)));
+    }
+
+    @Override
+    protected void preStart() {
+        setDockerImageName(System.getProperty(KEYCLOAK_PROPERTY, "quay.io/keycloak/keycloak:8.0.1"));
```
jstastny-cz: could we turn the image specs to a constant too? KEYCLOAK_IMAGE for example? so that it's in top of the class definition.

examples-test-utils/src/main/java/org/kie/kogito/testcontainers/quarkus/ConditionalQuarkusResource.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Optional;
+
+import org.kie.kogito.testcontainers.ConditionalGenericContainer;
+
+import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
+
+/**
+ * Quarkus resource to be run if and only if it was enabled.
+ */
+public abstract class ConditionalQuarkusResource implements QuarkusTestResourceLifecycleManager {
```
jstastny-cz: Since the class is called `ConditionalQuarkusResource` I would expect this to be a general use test resource, but it's tightly coupled with the containers, the name should reflect that too. Plus I would also keep the full substring `TestResource` as part of the name.

Wouldn't here be more general approach to extend the `QUarkusTestResourceLifecycleManager` interface by a method like `boolean isEnabled()` that would then be used in start/stop. That we can call ConditionalTestResource and then by all means having an abstract class that introduces the container support to be extended by particular testcontainers based test resources.

examples-test-utils/src/main/java/org/kie/kogito/testcontainers/quarkus/KeycloakQuarkusResource.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.kie.kogito.testcontainers.quarkus;
+
+import org.kie.kogito.testcontainers.KeycloakContainer;
+
+/**
+ * Keycloak quarkus resource that works within the test lifecycle.
+ *
+ */
+public class KeycloakQuarkusResource extends ConditionalQuarkusResource {
```
jstastny-cz: I would keep the word `TestResource` in names.

jstastny-cz (changes requested):

examples-test-utils/src/main/java/org/kie/kogito/testcontainers/KeycloakContainer.java
Sgitario: Sorry, this is a leftover. I will remove it instead.

examples-test-utils/src/main/java/org/kie/kogito/testcontainers/quarkus/ConditionalQuarkusResource.java
Sgitario: If I understood correctly your comment, you are proposing either (a) to rename this class with a better name as this is tightly coupled with testcontainers, or (b) isolate it from testcontainers.

I partly did (b). Ideally, the ConditionalGenericContainer class should be isolated as well, but this would not be compatible with the builder pattern: doing this.enableIfTestCategoryIs().withEnv("x", "y), and for spring boot, I think this is quite useful as it allows to set your container in one line. 

Let me know if you're ok with these changes. Thanks!

examples-test-utils/src/main/java/org/kie/kogito/testcontainers/quarkus/KeycloakQuarkusResource.java
Sgitario: Renamed this and InfinispanQuarkusResource.

examples-test-utils/src/main/java/org/kie/kogito/resources/ConditionalTestResource.java
```diff
@@ -0,0 +1,41 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.kie.kogito.resources;
+
+/**
+ * Conditional test resource for kogito tests.
+ * 
+ */
+public interface ConditionalTestResource<T> {
```
jstastny-cz, line 22: I meant extending `QuarkusTestResourceLifecycleManager` here and implementing just this in `ConditionalQuarkusTestResource`
`public abstract class ConditionalQuarkusTestResource implements ConditionalTestResource`.

Sgitario: As agreed, I've already renamed the field to conditionalResource. Thanks for spotting this!

process-usertasks-with-security-oidc-springboot/src/test/java/org/acme/travels/ApprovalsRestIT.java
```diff
@@ -22,41 +22,42 @@
...
+@Testcontainers
+@ExtendWith(SpringExtension.class)
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT, classes = KogitoSpringbootApplication.class)
 @DirtiesContext(classMode = ClassMode.AFTER_EACH_TEST_METHOD)
 public class ApprovalsRestIT {
 
-    @ClassRule
-    public static final KeycloakContainerResource keycloak = new KeycloakContainerResource();
+    @Container
+    public static KeycloakContainer KEYCLOAK = new KeycloakContainer();
```
jstastny-cz: can be still final, right?

jstastny-cz (approved):
Just a small nitpick, otherwise looks good.

process-usertasks-with-security-oidc-springboot/src/test/java/org/acme/travels/ApprovalsRestIT.java
Sgitario: Sure. Re-added.

PR #375, opened 2020-09-30
KOGITO-3480: Native Tests are not being executed in Kogito Examples

Sgitario:
JIRA: https://issues.redhat.com/browse/KOGITO-3480
This task scopes:
1. Enable the execution of Native tests using the native maven profile:

```java
mvn clean install -Pnative
```

2. Fix native tests failing. Common issues:

Many thanks for submitting your Pull Request :heart:! 

Please make sure that your PR meets the following requirements:

**WARNING! Please make sure you are opening your PR against `master` branch!**

- [x] You have read the [contributors guide](https://github.com/kiegroup/kogito-runtimes#contributing-to-kogito)
- [x] Pull Request title is properly formatted: `KOGITO-XYZ Subject`
- [x] Pull Request title contains the target branch if not targeting master: `[0.9.x] KOGITO-XYZ Subject`
- [x] Pull Request contains link to the JIRA issue
- [x] Pull Request contains link to any dependent or related Pull Request
- [x] Pull Request contains description of the issue
- [x] Pull Request does not include fixes for issues other than the main ticket

### DIFF
--- a/dmn-tracing-quarkus/src/test/java/org/kie/dmn/kogito/quarkus/tracing/LoanEligibilityTest.java
+++ b/dmn-tracing-quarkus/src/test/java/org/kie/dmn/kogito/quarkus/tracing/LoanEligibilityTest.java
@@ -15,17 +15,16 @@
  */
 package org.kie.dmn.kogito.quarkus.tracing;
 
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
-import javax.inject.Inject;
-
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import io.quarkus.test.common.QuarkusTestResource;
-import io.quarkus.test.junit.QuarkusTest;
-import io.restassured.http.ContentType;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.kie.kogito.kafka.KafkaClient;
 import org.kie.kogito.testcontainers.quarkus.KafkaQuarkusTestResource;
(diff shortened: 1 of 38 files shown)


dmn-tracing-quarkus/src/test/java/org/kie/dmn/kogito/quarkus/tracing/LoanEligibilityTest.java
```diff
@@ -15,55 +15,57 @@
...
 
-    @Test
-    public void testEvaluateTrafficViolation() throws InterruptedException {
+    @BeforeEach
+    public void setup() {
         kafkaClient = new KafkaClient(kafkaBootstrapServers);
+    }
 
+    @Test
+    public void testEvaluateTrafficViolation() throws InterruptedException {
```
tarilabs: Added @r00ta since the changes in this test besides of the other pom changes.

MarianMacik (approved):

pom.xml
```diff
@@ -232,6 +233,74 @@
         <module>process-springboot-example</module>
       </modules>
     </profile>
+
+    <profile>
+      <id>native</id>
+      <properties>
+        <failsafe.include>**/Native*IT.java</failsafe.include>
+        <failsafe.exclude></failsafe.exclude>
```
tarilabs, line 243:
```suggestion
        <failsafe.include>**/*IT.java,**/IT*.java,**/*ITCase.java</failsafe.include>
        <failsafe.exclude></failsafe.exclude>
```

wouldn't it be simpler to restore it as default, since native test are anyway annotated?

cristianonicolai (approved):

r00ta (approved):
LGTM thanks for the ping!

pom.xml
Sgitario: Do you mean to not set the "failsafe.include" property (as the default one "**/*IT.java" already satisfy the native tests? If so, I'm fine with that, I will do it. But using the regular expressions that you suggest seems way more complex.

tarilabs: I meant to restore the Failsafe's default: https://maven.apache.org/surefire/maven-failsafe-plugin/examples/inclusion-exclusion.html#Inclusions

Sgitario: This won't work as Quarkus needs a special annotation @NativeImageTest which only has it the Native*IT.java classes to work. If a IT test is executed in native mode, it won't work.

tarilabs: > This won't work as Quarkus needs a special annotation @NativeImageTest which only has it the Native*IT.java classes to work

interesting, I didn't know that. Can you point me to the reference where the annotation enforces the name to be `Native*IT.java`, please?

danielezonca: I had a look to quarkus doc because I remember something similar but the required setup for native testing seems to be a bit different
https://quarkus.io/guides/getting-started-testing#native-executable-testing
https://quarkus.io/guides/building-native-image#testing-the-native-executable
I don't see any specific filter on resources

@Sgitario can you please verify?

Sgitario: Let me be clearer on this:

- The @NativeImageTest annotation from Quarkus does not require that the class name follows "Native*IT, we do. I meant that all the Native*IT.java classes we have in Kogito Examples, use this annotation (and that's why we are safe to use this regular expression in the pom.xml). 

- We need to set the property "failsafe.include" to "**/Native*IT.java" to not execute the *IT classes. Otherwise, when there are both *IT.java and Native*IT.java classes, the build fails at integration-test stage (the *IT test fails). You can try yourself.

However, let me know if you think something here can be improved or done differently.

Sgitario:
Jenkis retest this

Sgitario:
Jenkins test this

r00ta:
Since these tests are included in many pipelines and the native tests are really time consuming, do we have an estimation on the impact @Sgitario?

Sgitario:
> Since these tests are included in many pipelines and the native tests are really time consuming, do we have an estimation on the impact @Sgitario?

We're not going to run the native profile in the pipelines (it's really time consuming).

danielezonca:
@r00ta 
We plan to add a [daily job](https://github.com/kiegroup/kogito-runtimes/pull/793) to execute native build + tests

r00ta:
Sounds good, thanks @danielezonca @Sgitario

pom.xml
danielezonca: Ok thanks for your explanation. I see your point, we have some ITs that requires native mode and others not so we decided to apply this naming convention to solve this.
I think we should better document it but it is out of scope for this PR :)

danielezonca (approved):

pom.xml
```diff
@@ -232,6 +233,74 @@
...
+        <module>process-usertasks-custom-lifecycle-quarkus</module>
+        <module>process-usertasks-quarkus</module>
+        <module>process-usertasks-with-security-oidc-quarkus</module>
+        <module>process-usertasks-with-security-quarkus</module>
+        <module>rules-quarkus-helloworld</module>
+        <module>ruleunit-quarkus-example</module>
+        <module>serverless-workflow-greeting-quarkus</module>
+        <module>serverless-workflow-service-calls-quarkus</module>
+        <module>serverless-workflow-events-quarkus</module>
+      </modules>
```
danielezonca, line 284: One question, why just don't apply this configuration to all the modules instead of enumerate the modules where we can have native tests?

Sgitario: This is to group only the quarkus modules and hence exclude the spring boot examples.

tarilabs (approved):
thanks!
Approved with the same minor comment as Daniele here https://github.com/kiegroup/kogito-examples/pull/375/commits/5e23bf8f924ea6651a9d01b9bf78f6459072fad1#r498134919

or at least have a enforcer/sonarcloud check if some `Test` or `IT` was added but for the naming convention is potentially excluded (doesn't have to be this PR, can be a separate JIRA) wdyt?

manstis:
@Sgitario FYI, see https://github.com/kiegroup/kogito-runtimes/pull/801

Sgitario:
It failed because https://issues.redhat.com/browse/KOGITO-3541. Therefore, this PR depends on https://github.com/kiegroup/kogito-apps/pull/472.

Sgitario:
Jenkins retest this

cristianonicolai:
@Sgitario needs to fix a conflict now

PR #431, opened 2020-11-03
KOGITO-3753 CloudEvents: Update Documentation

evacchi:
https://issues.redhat.com/browse/KOGITO-3753

tiny change to the README of travel agency example


radtriste (approved):

ricardozanini (approved):

cristianonicolai:
@evacchi this used to trigger different paths in the process, is that still the case? is that going to the same topic but different channel?

evacchi:
same topic but different message type so that should still work

PR #446, opened 2020-11-09
[KOGITO-3771] Drools -Support accumulate with native image #3218

lucamolteni:
Test for https://github.com/kiegroup/drools/pull/3218

### DIFF
--- a/ruleunit-quarkus-example/src/main/java/org/kie/kogito/queries/AllAmounts.java
+++ b/ruleunit-quarkus-example/src/main/java/org/kie/kogito/queries/AllAmounts.java
@@ -1,0 +1,32 @@
+/**
+ *  Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


tkobayas:
I'm a little wondering about how we add a test to kogito when drools is fixed. Currently `kogito-example` master depends on `org.drools:drools-model-compiler:jar:7.46.0.t20201030` so we can manually reproduce the issue with the `/find-all-application-amounts` added by this PR but it's not tested by CI with native-build because `ruleunit-quarkus-example` doesn't have a unit test with `@NativeImageTest`, correct? Then, if we add a `@NativeImageTest` test, it will break the CI because `kogito-example` master doesn't yet consume drools 7.46.0-SNAPSHOT. The same is applied to integration-tests in `kogito-runtimes`. Sorry if this is something already discussed...

lucamolteni:
@tkobayas that's correct, I didn't think about it

I think I should add the test with @NativeImageTest in the Kogito repository directly and merge it when we bump the version.
In the meanwhile we should be able to merge this, as it's just an example for accumulate...

mariofusco (approved):
Now that kogito-runtimes depends on kie 7.46.0.Beta1 I think that you can also add a test for the native image generation. For the rest that's ok for me.

tkobayas (approved):

danielezonca (approved):
LGTM :+1:

hellowdan (approved):

lucamolteni:
Dockergeddon

```
[ERROR] 	... 6 more
[ERROR] Caused by: com.google.api.client.http.HttpResponseException: 429 Too Many Requests
[ERROR] {
[ERROR]   "errors": [
[ERROR]     {
[ERROR]       "code": "TOOMANYREQUESTS",
[ERROR]       "message": "You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit"
[ERROR]     }
[ERROR]   ]
[ERROR] }
```

mariofusco:
jenkins test

PR #541, opened 2021-01-30
KOGITO-3789: Update examples to visualize svg

nmirasch:
https://issues.redhat.com/browse/KOGITO-3789

Modified docker-compose configuration at kogito-travel-agency and quickstarters - 01 example to be able to visualize process diagrams at management console.

### DIFF
--- a/kogito-quickstarts/01-kogito-basic-example/docker-compose/startServices.sh
+++ b/kogito-quickstarts/01-kogito-basic-example/docker-compose/startServices.sh
@@ -45,5 +45,10 @@ else
     exit 1
 fi
 
-export DOCKER_GATEWAY_HOST=172.17.0.1
+if [ "$(uname)" == "Darwin" ]; then
+    export DOCKER_GATEWAY_HOST=$host.docker.internal
+elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+   export DOCKER_GATEWAY_HOST=172.17.0.1
+fi
+
 docker-compose up
--- a/kogito-quickstarts/01-kogito-basic-example/docker-compose/startServices.sh
+++ b/kogito-quickstarts/01-kogito-basic-example/docker-compose/startServices.sh
@@ -45,5 +45,5 @@ else
     exit 1
 fi
 
-
+export DOCKER_GATEWAY_HOST=172.17.0.1
 docker-compose up
--- a/kogito-quickstarts/01-kogito-basic-example/docker-compose/startServices.sh
+++ b/kogito-quickstarts/01-kogito-basic-example/docker-compose/startServices.sh
@@ -16,6 +16,12 @@ fi
 echo "Kogito Image version: ${KOGITO_VERSION}"
 echo "KOGITO_VERSION=${KOGITO_VERSION}" > ".env"
 
+if [ "$(uname)" == "Darwin" ]; then
+   echo "DOCKER_GATEWAY_HOST=${host.docker.internal}" >> ".env"
+elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+   echo "DOCKER_GATEWAY_HOST=172.17.0.1" >> ".env"
+fi
+
 PERSISTENCE_FOLDER=./persistence
 KOGITO_EXAMPLE_PERSISTENCE=../target/classes/META-INF/resources/persistence/protobuf
 
@@ -43,12 +49,6 @@ then
 else
     echo "$KOGITO_HIRING_SVG_FOLDER does not exist. Have you compiled 01-kogito-basic-example project?"
     exit 1
-fi
-
-if [ "$(uname)" == "Darwin" ]; then
-    export DOCKER_GATEWAY_HOST=$host.docker.internal
-elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
-   export DOCKER_GATEWAY_HOST=172.17.0.1
 fi
 
 docker-compose up
--- a/kogito-quickstarts/01-kogito-basic-example/docker-compose/startServices.sh
+++ b/kogito-quickstarts/01-kogito-basic-example/docker-compose/startServices.sh
@@ -31,4 +31,19 @@ else
     exit 1
 fi
 
+SVG_FOLDER=./svg
+
+KOGITO_HIRING_SVG_FOLDER=../target/classes/META-INF/processSVG
+
+mkdir -p $SVG_FOLDER
+
+if [ -d "$KOGITO_HIRING_SVG_FOLDER" ]
+then
+    cp $KOGITO_HIRING_SVG_FOLDER/*.svg $SVG_FOLDER
+else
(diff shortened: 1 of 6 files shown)


kogito-travel-agency/extended/docker-compose/docker-compose.yml
```diff
@@ -91,8 +91,13 @@ services:
 
   management-console:
     container_name: management-console
-    image: quay.io/kiegroup/kogito-management-console:${KOGITO_VERSION}
+    network_mode: "host"
+    image: quay.io/kiegroup/kogito-management-console-nightly:${KOGITO_VERSION}
     ports:
-      - 8280:8080
+      - 8280:8280
```
pefernan: You don't need *ports* when using *network_mode: "host"*

pefernan (changes requested):
Works great, the only minor thing is that I'd remove the **management-console** **ports** from the docker-compose file since it's not needed in network_mode **host**

pefernan:
A side comment, I think we should enhance the consoles docker images to allow configure the HTTP port with a specific environment variable. I had a similar issue that made me use a profile to change the port in another quickstart.

nmirasch:
@pefernan Removed ports mapping.
About that port configuration.. I have tried to change directly that 'quarkus.http.port' but only achieved to make it work modifying that inside a quarkus.profile.. We can explore that and improve this implementation if we'll find better solution, but for now I think make sense to provide a way to see the images, Wdyt?

pefernan:
@nmirasch yea, the docker image script is hardcoding the _quarkus.http.port_ to **8080** after passing the _KOGITO_MANAGEMENT_CONSOLE_PROPS_ so it overrides whatever you pass there. We can go with profile for now, and we we may improve the consoles scripts to have a kind of _HTTP_PORT_ env that sets the `quarkus.http.port` (defaulting to 8080), in a similar way as they do with the data index url.

pefernan (approved):

nmirasch:
@cristianonicolai @pefernan
I have added the svg to kogito travel agency and getting starters examples.. Removed network_mode: host and replaced by ${DOCKER_GATEWAY_HOST:-host.docker.internal} at the configuration refering to localhost. 
Could you take a look, please? Thanks!!

kogito-travel-agency/extended/docker-compose/docker-compose.yml
```diff
@@ -91,8 +91,13 @@ services:
 
   management-console:
     container_name: management-console
-    image: quay.io/kiegroup/kogito-management-console:${KOGITO_VERSION}
+    image: quay.io/kiegroup/kogito-management-console-nightly:${KOGITO_VERSION}
```
cristianonicolai: I nightly builds can be used for testing but not the committed version, the stable branch should have a set version to use.

kogito-quickstarts/01-kogito-basic-example/docker-compose/startServices.sh
```diff
@@ -31,4 +31,19 @@ else
...
+
+if [ -d "$KOGITO_HIRING_SVG_FOLDER" ]
+then
+    cp $KOGITO_HIRING_SVG_FOLDER/*.svg $SVG_FOLDER
+else
+    echo "$KOGITO_HIRING_SVG_FOLDER does not exist. Have you compiled 01-kogito-basic-example project?"
+    exit 1
+fi
+
+export DOCKER_GATEWAY_HOST=172.17.0.1
```
cristianonicolai: I like this idea but a sh script can be used for both Mac or Linux, so my suggestion would be to tweak it here to depending on the OS, send this IP or the host.docker.internal. See for instance https://stackoverflow.com/questions/3466166/how-to-check-if-running-in-cygwin-mac-or-linux

kogito-travel-agency/extended/docker-compose/docker-compose.yml
nmirasch: Removed the -nightly reference from docker images

kogito-quickstarts/01-kogito-basic-example/docker-compose/startServices.sh
nmirasch: @cristianonicolai I have added that DOCKER_GATEWAY_HOST env var setting depending on the SO at start-services.sh, but I couldn't try this on MAC, Could you take a look here and see if this fits?

kogito-travel-agency/extended/docker-compose/startServices.sh
```diff
@@ -53,4 +53,33 @@ else
...
+if [ -d "$KOGITO_VISAS_SVG_FOLDER" ]
+then
+    cp $KOGITO_VISAS_SVG_FOLDER/*.svg $SVG_FOLDER
+else
+    echo "$KOGITO_VISAS_SVG_FOLDER does not exist. Have you compiled Kogito Visas project?"
+    exit 1
+fi
+
+if [ "$(uname)" == "Darwin" ]; then
+    export DOCKER_GATEWAY_HOST=$host.docker.internal
```
pefernan: Nitpicking here but for consistency, would it be possible to have this env in the `.env` file as done for the [KOGITO_VERSION](https://github.com/kiegroup/kogito-examples/blob/84fa81c309037564831bf3183d3840017b95c178/kogito-travel-agency/extended/docker-compose/startServices.sh#L32)?

Or make the `KOGITO_VERSION` an export :D

pefernan (changes requested):
@nmirasch cool! But just a minor comment

kogito-travel-agency/extended/docker-compose/docker-compose.yml
pefernan: I personally think that would be good to use nightly when using the examples from master. Could we check if version ends with `.Final` to decide using nightly or not?

But this is not blocking for me, I can go withouth nightly :D

nmirasch:
@pefernan I've replaced the export by the .env addition.. could you take a look?

pefernan (approved):
+1 Thanks @nmirasch

cristianonicolai (approved):

jstastny-cz:
The jira description mentions 

> Update kogito travel agency example (scripts and instructions) 

Did we review the instructions for affected examples that there's no need to change?

nmirasch:
@jstastny-cz  This PR will cover this jira https://issues.redhat.com/browse/KOGITO-4228 too. Regarding we are applying the same solution to see the process diagram in the different examples. I can rename the commit and the PR to include  KOGITO-4228,Does it makes sense?

jstastny-cz:
@nmirasch what was result of our discussion? :-) do we want to add instructions here or as a follow-up?

nmirasch:
@jstastny-cz  I think the instructions about how to add the diagram visualization are detailed as part as the documentation This code is just to configure  show cases, I guess this  would be enough: the example + the instructions at the official doc. I would merge this PR to be able to see the diagrams at the current examples.

jstastny-cz (approved):

PR #542, opened 2021-02-01
KOGITO-4202: Quickstart index.html

Sara4994:
Hello,

Please review this PR intended to add UI(index.html) on kogito-quickstart with some quick links 

JIRA issue: https://issues.redhat.com/browse/KOGITO-4202

### DIFF
--- a/kogito-quickstarts/01-kogito-basic-example/src/main/resources/META-INF/resources/index.html
+++ b/kogito-quickstarts/01-kogito-basic-example/src/main/resources/META-INF/resources/index.html
@@ -137,7 +137,7 @@
       <section style="height: 350px; margin-top: 24px;">
           <div class="container">
               <div class="row">
-                  <h2> Quick Links</h1>
+                  <h2> Quick Links</h2>
                   <ul style="list-style: none;">
                       <li style="line-height: 40px;"><a href="https://kogito.kie.org/get-started/">Documentation</a></li>
                       <li style="line-height: 40px;"><a href="https://kie.zulipchat.com/#narrow/stream/232659-general">Getting Help</a></li>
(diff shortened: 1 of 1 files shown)


nmirasch:
@Sara4994 for the 02 quickstart this index don't work regarding authentication is not enabled as a web-app. I would suggest just bring this index.html to the 01 basic quick start

kogito-quickstarts/01-kogito-basic-example/src/main/resources/templates/index.html
```diff
@@ -0,0 +1,150 @@
...
+                          Latest updates
+                      </a>
+                  </div>
+              </div>
+          </div>
+      </section>
+      <section style="height: 350px; margin-top: 24px;">
+          <div class="container">
+              <div class="row">
+                  <h2> Quick Links</h1>
```
nmirasch: @Sara4994 this is a bit strange :-)

nmirasch (commented):
@Sara4994 After taking a look there I don't see the point of use quarkus-qute when there is only static content. I think is better  to add just the index.html directly to resources/META-INF/resources/index.html

Sara4994:
> @Sara4994 After taking a look there I don't see the point of use quarkus-qute when there is only static content. I think is better to add just the index.html directly to resources/META-INF/resources/index.html

@nmirasch done, I have removed qute and just placed index.html under resources folder

nmirasch:
@Sara4994 looks good now.  I have added this index.html as part of the movement to examples https://github.com/kiegroup/kogito-examples/pull/599. I prefer not to merge this one now regarding it will be included in the updated examples when that PR is merged. I ask you to keep this PR open until the movement have been done, Thanks again!

PR #581, opened 2021-02-19
[1.3.x] KOGITO-4505 Use legacy-jar packaging in all Quarkus examples

evacchi:
https://issues.redhat.com/browse/KOGITO-4505
images/operators are still using the legacy file name


evacchi:
probably not necessary, we're downgrading

PR #582, opened 2021-02-19
KOGITO-4505 Use legacy-jar packaging in all Quarkus examples

evacchi:
https://issues.redhat.com/browse/KOGITO-4505
work-around to images/operators expecting the old name

PR #594, opened 2021-03-01
Revert "KOGITO-4505 Use legacy-jar packaging in all Quarkus examples"

spolti:
Reverts kiegroup/kogito-examples#582
Issue fixed on images as part of https://issues.redhat.com/browse/KOGITO-4503

### DIFF
--- a/decisiontable-quarkus-example/src/main/resources/application.properties
+++ b/decisiontable-quarkus-example/src/main/resources/application.properties
@@ -1,6 +1,3 @@
-# Packaging
-quarkus.package.type=legacy-jar
-
 quarkus.swagger-ui.always-include=true
 
 # Maximum Java heap to be used during the native image generation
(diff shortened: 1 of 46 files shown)


spolti:
probably it failed because I reverted the the changes using the previous PR's revert button.
@radtriste is there a way to make the tests starts without redo the changes on my fork?

spolti:
depends on https://github.com/kiegroup/kogito-images/pull/378

cristianonicolai:
@spolti I think we might need to hold this as master will be reverted to 1.11, right @evacchi?

evacchi:
I think it may be worth to do the opposite change and use fast-jar explicitly (forward compatibility), WDYT @cristianonicolai ?

cristianonicolai:
> I think it may be worth to do the opposite change and use fast-jar explicitly (forward compatibility), WDYT @cristianonicolai ?

makes sense, +1

radtriste:
@spolti @evacchi I guess this one can be close as we are moving forward with https://github.com/kiegroup/kogito-examples/pull/601

PR #595, opened 2021-03-01
[KOGITO-4503] - Update to Quarkus 1.12 on artifacts broke the images

spolti:
Many thanks for submitting your Pull Request :heart:! 
See https://issues.redhat.com/browse/KOGITO-4503

Depends on https://github.com/kiegroup/kogito-examples/pull/594

Please make sure that your PR meets the following requirements:

**WARNING! Please make sure you are opening your PR against `master` branch!**

- [x] You have read the [contributors guide](https://github.com/kiegroup/kogito-runtimes#contributing-to-kogito)
- [x] Pull Request title is properly formatted: `KOGITO-XYZ Subject`
- [x] Pull Request title contains the target branch if not targeting master: `[0.9.x] KOGITO-XYZ Subject`
- [x] Pull Request contains link to the JIRA issue
- [x] Pull Request contains link to any dependent or related Pull Request
- [x] Pull Request contains description of the issue
- [x] Pull Request does not include fixes for issues other than the main ticket

PR #601, opened 2021-03-05
KOGITO-4595 Downgrade Quarkus 1.11 -- use fast-jar everywhere

evacchi:
https://github.com/kiegroup/kogito-runtimes/pull/1115
https://github.com/kiegroup/kogito-apps/pull/692
https://github.com/kiegroup/kogito-examples/pull/601

https://issues.redhat.com/browse/KOGITO-4595


dmn-drools-quarkus-metrics/Dockerfile
```diff
@@ -1,4 +1,8 @@
-FROM quay.io/kiegroup/kogito-quarkus-jvm-ubi8:latest
+FROM quay.io/kiegroup/kogito-runtime-jvm:latest
+
+COPY target/quarkus-app/lib/ $KOGITO_HOME/bin/lib/
+COPY target/quarkus-app/*.jar $KOGITO_HOME/bin
+COPY target/quarkus-app/app/ $KOGITO_HOME/bin/app/
+COPY target/quarkus-app/quarkus/ $KOGITO_HOME/bin/quarkus/
+COPY scripts/kogito-app-launch.sh $KOGITO_HOME/kogito-app-launch.sh
```
evacchi: @r00ta @spolti @cristianonicolai FYI had to fix this launcher script as well

r00ta: Hi @evacchi , just for my understanding: since this is the "quickstart" repo for kogito users, are we suggesting to use `fast-jar` also in the user's application? If so, shouldn't we
1) generate the property with the kogito-codegen
2) fix the base image in  `kogito-images` so to keep the same usage in the Dockerfile?

nmirasch (approved):

dmn-drools-quarkus-metrics/Dockerfile
evacchi: good point for 1) for 2) @spolti is on it

radtriste: raised also https://kie.zulipchat.com/#narrow/stream/232676-kogito/topic/Examples.20testing.20using.20Kogito.20images.20bring.20cyclic.20dependency as Dockerfiles should not be used by testing (else would make a cyclic dependency with images ...)

radtriste: and https://issues.redhat.com/browse/KOGITO-4609 should solve some local build if needed on master

cristianonicolai (approved):

dmn-drools-quarkus-metrics/Dockerfile
```diff
@@ -2,5 +2,9 @@ FROM quay.io/kiegroup/kogito-runtime-jvm:latest
 
 ENV RUNTIME_TYPE quarkus
 
-COPY target/*-runner.jar $KOGITO_HOME/bin
-COPY target/lib $KOGITO_HOME/bin/lib
+COPY target/quarkus-app/lib/ $KOGITO_HOME/bin/lib/
```
radtriste: With the merge of https://github.com/kiegroup/kogito-examples/commit/2cd1582a544be06b54d4b4514727d2c04b33c848, this should not be needed anymore in that PR and will be handled by https://github.com/kiegroup/kogito-examples/pull/595

evacchi: what about the script runner?

radtriste: same it should be handled directly into the image

radtriste: See https://github.com/kiegroup/kogito-images/pull/378/files#diff-e40a5f5f48e7c9facc94ca62fd67bc804e7866908ca0091ec042620dabdb0fdb

evacchi: I am confused, the Dockerfile on master looks unchanged https://github.com/kiegroup/kogito-examples/blob/master/dmn-drools-quarkus-metrics/Dockerfile

evacchi: I am confused https://github.com/kiegroup/kogito-examples/commit/2cd1582a544be06b54d4b4514727d2c04b33c848 hasn't changed this Dockerfile, has it?

evacchi: Ok I suppose I have figured this out, but I thought we were upgrading everything to fast-jar; now this is the only example that uses legacy. It's ok, but it's probably better to upgrade it /cc @r00ta

r00ta: Hi @evacchi , now the example is not using anymore `dmn-drools-quarkus-metrics/Dockerfile` to build the docker image, so we don't have anymore the cyclic dependency. 

Given that this example should tell user how to build a kogito docker image, I suggest that `dmn-drools-quarkus-metrics/Dockerfile` should be always up to date with the changes in `kogito-images`. This `Dockerfile` should not contain custom/fancy script, it should always build the docker image as we suggest to do in the official kogito docs (these dockerfiles just to be explicit https://docs.jboss.org/kogito/release/latest/html_single/#proc-kogito-deploying-on-kubernetes_kogito-deploying-on-openshift)

evacchi: @r00ta the project had been re-configured to use fast-jar in application.conf, that is why the Dockerfile was being updated. This is to be forward-compatible with newer versions of Quarkus.  Nothing fancy is going on. The reason the script was being updated is that the builder image was not working at the time of testing (i.e. it was looking for the legacy-jar path instead of fast-jar's). I'll keep this as a "legacy" jar for now, but it's probably worth upgrading to fast-jar to align with the rest of the examples.

radtriste:
Jenkins retest this

dmn-drools-quarkus-metrics/Dockerfile
r00ta: as per chat I think we sorted it out :)

PR #605, opened 2021-03-09
KOGITO-4623 Setting fast-jar in application.conf breaks native build

evacchi:
Unset fast-jar in application.conf, use Maven property (fixes native build)


radtriste (commented):
Seems like some examples do not have the `<quarkus.package.type>native</quarkus.package.type>` in native profile. 
can you also add this ?

evacchi:
it's in the root pom @radtriste

radtriste:
a bit fuzzy those days ... sorry. 
anyway, we need to clean up a bit the examples I guess because some examples do repeat some configuration set in main pom

radtriste (approved):

PR #620, opened 2021-03-18
[KOGITO-4664] Add PMMLs with scorecard and mining model to examples

jiripetrlik:


gitgabrio (commented):
@jiripetrlik Thanks for that PR. I'm wondering if it would be better to also add "wrong data" tests: wrong data may be different on a by-model base, so covering them could give a better monitor on what is going on. WDYT ?

dmn-pmml-quarkus-example/src/test/java/org/kie/pmml/kogito/quarkus/example/NativeDecisionTreeIT.java
```diff
@@ -18,7 +18,7 @@
 import io.quarkus.test.junit.NativeImageTest;
 
 @NativeImageTest
-public class NativeSampleMineIT extends SampleMineTest {
+public class NativeDecisionTreeIT extends DecisionTreeTest {
```
danielezonca, line 21: Native test must follow `Native*IT` naming convention

jiripetrlik:
@gitgabrio Yes, you are right. I've added tests for wrong data to all models in Spring boot using previously developed tools.

gitgabrio (approved):
Thanks @jiripetrlik LGTM

danielezonca (approved):

PR #728, opened 2021-06-09
[KOGITO-5302] Disable fail on empty beans

fjtirado:
Just a quick fix for the example to work on native mode. 
We should consider to have a global customizer class where we set up that to false to all projects, but I let @danielezonca to think about this ;)

### DIFF
--- a/process-quarkus-example/src/main/java/org/kie/kogito/examples/CustomizeObjectMapper.java
+++ b/process-quarkus-example/src/main/java/org/kie/kogito/examples/CustomizeObjectMapper.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


tiagodolphine (approved):

cristianonicolai (approved):

evacchi (approved):

evacchi:
we do have a global objectmapper so if we do plan to allow this, then it is doable @danielezonca what were your concerns?

radtriste (approved):

danielezonca:
> we do have a global objectmapper so if we do plan to allow this, then it is doable @danielezonca what were your concerns?

My main/only concern is that as far as I know, in native mode if a class has not been marked for reflection the marshalling will just skip that field during deserialisation and you get a `null`.
We already had in the past similar problem
https://github.com/kiegroup/kogito-runtimes/pull/1111
And the solution was to add the classes to serialisation (+ add some native test that verified the expected result in detail).

I'm not a Jackson expert (and even less with Jackson in native mode) but are we sure this `SerializationFeature.FAIL_ON_EMPTY_BEANS` property is a general solution for this problem? Does it cover scenarios where only part of the payload cannot be deserialised (like a field)?
If so I'm fine to move this fix to runtimes to be applied globally 👍

fjtirado:
@danielezonca 
Generated classes are being included automatically by Quarkus. Thats why other example with no empty beans are workings fine. As I already mentioned in Zulip, explicityly adding the empty bean to reflection does not fix anything, because this class is already there. So, we do NOT need to change KogitoAssetProcessor to add anything there
@evacchi 
Where is this global object mapper defined?

fjtirado:
I closed this one and opened https://github.com/kiegroup/kogito-runtimes/pull/1366 to disable fail on empty beans globally

PR #733, opened 2021-06-17
[KOGITO-5259] BPMN rest example

fjtirado:
Two examples to illustrate the usage of RestWorkItemHandler  were added:
- First one is a copy paste of process-service-rest-call-quarkus, but using RestWorkItemHandler rather than the programatic approach. 
- Second one is a copy paste of serverless-workflow-functions-quarkus, but using a BPMN process rather than a serverless workflow

I ignore springboot examples since I think the quarkus ones should be enough

Merge with https://github.com/kiegroup/kogito-runtimes/pull/1370


fjtirado:
jenkins retest this

process-service-rest-multi-call-quarkus/README.md
```diff
@@ -0,0 +1,99 @@
+# Kogito Serverless Workflow - Rest Example
```
ricardozanini:
```suggestion
# Kogito Process - Rest Example
```

process-service-rest-multi-call-quarkus/README.md
```diff
@@ -0,0 +1,99 @@
...
+  
+To run the generated native executable, generated in `target/`, execute
+
+```text
+./target/sw-quarkus-greeting-{version}-runner
+```
+
+### Submit a request
+
+The process can be executed by sending a request to http://localhost:8080/RESTExample'
```
ricardozanini:
```suggestion
The process can be executed by sending a request to [http://localhost:8080/RestExample](http://localhost:8080/RestExample)
```

process-service-rest-multi-call-quarkus/README.md
```diff
@@ -0,0 +1,99 @@
...
+}
+```
+
+Complete curl command can be found below:
+
+```text
+curl -X POST -H 'Content-Type:application/json' -H 'Accept:application/json' -d '{"inputNumbers" : {"numbers": [1,2,3,4,5,6,7,8,7]}}' http://localhost:8080/RestExample
+```
+
+curl response will be something like this, which includes field ```sum```, the result of multiplying each input number by 8 (that number might differ in your execution) and summing all of them:
```
ricardozanini:
```suggestion
curl response will be something like this, which includes field `sum`, the result of multiplying each input number by 8 (that number might differ in your execution) and summing all of them:
```

process-service-rest-multi-call-quarkus/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+        <version>${project.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+  <dependencies>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>kogito-quarkus</artifactId>
```
ricardozanini:
```suggestion
      <artifactId>kogito-quarkus-processes</artifactId>
```
I believe you won't need DMN nor DRL in this example?

process-service-rest-multi-call-quarkus/pom.xml
```diff
@@ -0,0 +1,101 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>kogito-rest-workitem</artifactId>
```
ricardozanini: How about adding this  Work Item to `kogito-quarkus-processes` extension? This way, users will have it ootb. :thinking:

process-service-rest-workitem-quarkus/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+      <artifactId>quarkus-rest-client</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-smallrye-fault-tolerance</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.smallrye</groupId>
+      <artifactId>smallrye-context-propagation-propagators-rxjava2</artifactId>
+      <version>1.0.7</version>
```
ricardozanini: Can you add this version to the Examples BOM?

ricardozanini (commented):
Pretty nice examples! :+1:

process-service-rest-multi-call-quarkus/pom.xml
fjtirado: Hmmm, it is a temptation I shared, @cristianonicolai wdyt?

process-service-rest-workitem-quarkus/pom.xml
fjtirado: I do not know why this is there, I will review (the dangers of copy paste)

fjtirado: Removed

fjtirado:
jenkins retest this

process-service-rest-multi-call-quarkus/pom.xml
fjtirado: After talking with @lazarotti, he reminded me that Rest Work Item Handler is not going to be included out of the box (we are promoting the service approach after all in Kogito), so the answer is no, we should not include this in process extension.

fjtirado:
jenkins please test this

ricardozanini (approved):

ricardozanini:
Jenkins retest this

pom.xml
```diff
@@ -162,7 +162,8 @@
         <module>process-scripts-springboot</module>
         <module>process-service-calls-quarkus</module>
         <module>process-service-calls-springboot</module>
-        <module>process-service-rest-call-quarkus</module>
+        <module>process-service-rest-multi-call-quarkus</module>
```
tiagodolphine: do we really need 2 modules for thes examples? can we have both processes in the same module?

process-service-rest-multi-call-quarkus/README.md
```diff
@@ -0,0 +1,99 @@
+# Kogito Process - Rest Example
```
tiagodolphine: It is important to mention the `.wid` file to allow the user to use the rest work item in the editor

process-service-rest-multi-call-quarkus/src/main/java/org/acme/numbers/NumbersResource.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @PostConstruct
+    void init() {
+        objectMapper = new ObjectMapper();
+        rand = new Random();
+    }
+
+    @GET
+    @Path("random")
+    public Response getRandom(@QueryParam("bound") @DefaultValue("10") int bound) {
```
tiagodolphine: You can just return an Integer instead of Response, in this way avoid the `fromNumber` method.

process-service-rest-workitem-quarkus/README.md
```diff
@@ -0,0 +1,152 @@
+# Process Service invocation with REST call
```
tiagodolphine: So, can't this example be in the same as the multi one? or even in the `process-service-rest-call-quarkus` to show 2 different approaches to do REST calls?

process-service-rest-workitem-quarkus/src/main/resources/META-INF/resources/index.html
```diff
@@ -0,0 +1,152 @@
+<!DOCTYPE html>
```
tiagodolphine: I think this can be removed

process-service-rest-workitem-quarkus/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Description
+
+A quickstart project that processes users in the system. It's main purpose is to to call external REST service
+to load a given user by its username.
+
+There are two ways to invoke a rest web service in Kogito: programmatically, as shown in this other [example](https://github.com/kiegroup/kogito-examples/tree/stable/process-service-rest-call-quarkus) or using RestWorkItemHandler, as in this example.
```
tiagodolphine: the same to comment about the `.wid` file

pom.xml
cristianonicolai: +1 I guess we can combine and so reference this as rest examples.

process-service-rest-multi-call-quarkus/README.md
fjtirado: Ok, Ill do that.

process-service-rest-multi-call-quarkus/src/main/java/org/acme/numbers/NumbersResource.java
fjtirado: I know, but I wanted this to be similar to other resources.

process-service-rest-workitem-quarkus/src/main/resources/META-INF/resources/index.html
fjtirado: yes, copy-paste issue

pom.xml
fjtirado: @cristianonicolai @tiagodolphine 
hmmm, the reason for having two examples is pretty simple (I should have indicated it more clearly in the PR description):  process-service-rest-workitem-quarkus is a copy of process-service-rest-call-quarkus and this intention is to show up the differences between our preferred approach to deal with rest invocations (the service approach) and the declarative alternative (the workitem approach). I was planning to use both examples in an upcoming blog entry to discuss when to use one or another. 
Since the first example does not shows up consecutive rest invocation, I feel a more complex, second example, was needed, hence the addition of process-service-rest-multi.class-quarkus, which illustrates how the output of the first GET invocation should be used in the second POST invocation. 
Combining both examples will kill their intended purpose, so I think they need to stay (I renamed the multi one to include the workitem suffix)

process-service-rest-multi-call-quarkus/src/main/java/org/acme/numbers/NumbersResource.java
fjtirado: Will change anyway

process-service-rest-workitem-quarkus/README.md
fjtirado: see https://github.com/kiegroup/kogito-examples/pull/733#discussion_r656031937

process-service-rest-workitem-quarkus/README.md
fjtirado: sure

pom.xml
cristianonicolai: I guess the point is more about having two processes inside the same project. Each bpmn could exemplify one scenario that you described or would that somehow limit what you want to showcase?

fjtirado: Ok, I see the point now. Thanks ;)
Actually I think grouping non dependant projects (project that can be executed separally and do not interact between them, but are conceptually related) will be kind of confusing and we are not doing it (for example, projects related with event consumption and publishing) . I mean, right now we are grouping examples, as far as I know (travel-agency, onboading...), when somehow the projects are dependent one on another.
 I see the benefits of grouping related functionality examples, but I feel this should be a different JIRA which applies to the whole repo, because if not, we will not be following a coherent approach, wdyt?

fjtirado: @ricardozanini wdyt? Should we group all rest example under the same project or should we keep them on the root level?

fjtirado: in the meantime, I changed the names of the projects, so all start with process-rest, I think this is a good alternative to grouping.

ricardozanini: This is tricky. I don't have a strong opinion about it. Normally, my rule of thumb is creating a multi-module project when the context brings them together or needs to be built together. In this case, they have the same concept but are not strongly related. I'm fine with both approaches (multi or separated projects).

fjtirado: @tiagodolphine I think as Ricardo, I believe we should group then only they dependent on each other, which is not the case. The renaming I made should be enough to conceptually "group" them by the user going though the examples (all rest related process for bpmn now start with process-rest)

ricardozanini:
@tiagodolphine are you ok with this PR?

pom.xml
tiagodolphine: Ok, that's fine keeping a prefix in the modules if we decide at some point to group similar modules this involves not just the rest examples, but others as well, so let's keep the same pattern we already have.

tiagodolphine (approved):

PR #828, opened 2021-08-03
KOGITO-5267 Update Quickstarts to allow user to override Quarkus BOM

antncl:
JIRA https://issues.redhat.com/browse/KOGITO-5267

Updating the Quickstarts to allow the user to override the Quarkus BOM. Intentionally avoided using `${version.io.quarkus}` in place of the static versions as 1) the example in the ticket doesn't use it 2) PME on the prod side would not override this property and 3) its possible the platform would be rebuild multiple times and these two properties would not be the same (?).

If others are OK with the above we should probably devise some automation to update these versions too. Maybe enhancement based on @mbiarnes automation that updates the snapshot versions.

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>Pull Request</b>  
  Please add comment: <b>Jenkins retest this</b>
 
* <b>Quarkus LTS checks</b>  
  Please add comment: <b>Jenkins run LTS</b>

* <b>Native checks</b>  
  Please add comment: <b>Jenkins run native</b>

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,18 @@ This module contains a number of examples that you can take a look at and try ou
 
 Since Kogito aims at supporting both Quarkus and Spring Boot each example usually provides both type of projects.
 
-- Default branch is `stable`, pointing to the latest released version. 
+- Default branch is `stable`, pointing to the latest released version.
 - **[You can also check all versions by looking at releases.](https://github.com/kiegroup/kogito-examples/releases/latest)**
+
+## Use alternative Quarkus platforms
+
+The Quarkus quickstarts by default currently use the Quarkus core BOM.
+
+If you want to use an alternative BOM when building the Quarkus quickstarts you can override the `quarkus.platform.*` properties. The following example shows how to set `quarkus.platform.artifact-id` to use the quarkus-universe-bom.
+
+```
+mvn -Dquarkus.platform.artifact-id=quarkus-universe-bom clean install
+```
 
 
 ## Contribution
@@ -113,7 +123,7 @@ of how to install the operator to your environment in order to try it there.
 ### Issues
 - Do you have a [minimal, reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) for your issue?
   - If so, please open a Jira for it in the [Kogito project](https://issues.redhat.com/projects/KOGITO/summary) with the details of your issue and example.
-- Are you encountering an issue but unsure of what is going on? 
+- Are you encountering an issue but unsure of what is going on?
   - Start a new conversation in the Kogito [Google Group](https://groups.google.com/g/kogito-development), or open a new thread in the [Kogito stream](https://kie.zulipchat.com/#narrow/stream/232676-kogito) of the KIE Zulip chat.
   - Please provide as much relevant information as you can as to what could be causing the issue, and our developers will help you figure out what's going wrong.
 
--- a/decisiontable-quarkus-example/pom.xml
+++ b/decisiontable-quarkus-example/pom.xml
@@ -9,12 +9,18 @@
   </parent>
   <artifactId>decisiontable-quarkus-example</artifactId>
   <name>Kogito Example :: Decision Table - Quarkus</name>
+  <properties>
+    <quarkus-plugin.version>2.1.0.Final</quarkus-plugin.version>
+    <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
+    <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
+    <quarkus.platform.version>2.1.0.Final</quarkus.platform.version>
+  </properties>
   <dependencyManagement>
     <dependencies>
       <dependency>
-        <groupId>org.kie.kogito</groupId>
-        <artifactId>kogito-quarkus-bom</artifactId>
-        <version>${project.version}</version>
+        <groupId>${quarkus.platform.group-id}</groupId>
+        <artifactId>${quarkus.platform.artifact-id}</artifactId>
+        <version>${quarkus.platform.version}</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
@@ -70,6 +76,7 @@
       <plugin>
         <groupId>io.quarkus</groupId>
         <artifactId>quarkus-maven-plugin</artifactId>
+        <version>${quarkus-plugin.version}</version>
         <executions>
           <execution>
             <goals>
--- a/dmn-drools-quarkus-metrics/pom.xml
+++ b/dmn-drools-quarkus-metrics/pom.xml
@@ -9,12 +9,18 @@
   </parent>
   <artifactId>dmn-drools-quarkus-metrics</artifactId>
   <name>Kogito Example :: DMN Metrics Quarkus</name>
+  <properties>
+    <quarkus-plugin.version>2.1.0.Final</quarkus-plugin.version>
+    <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
+    <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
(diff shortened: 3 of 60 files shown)


mareknovotny (approved):
it looks good to me

antncl:
Unrelated failures afaict

antncl:
jenkins retest this

antncl:
@danielezonca would you take a look at these changes please? thank you!

decisiontable-quarkus-example/pom.xml
```diff
@@ -9,12 +9,18 @@
   </parent>
   <artifactId>decisiontable-quarkus-example</artifactId>
   <name>Kogito Example :: Decision Table - Quarkus</name>
+  <properties>
+    <quarkus-plugin.version>2.1.0.Final</quarkus-plugin.version>
+    <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
+    <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
+    <quarkus.platform.version>2.1.0.Final</quarkus.platform.version>
```
danielezonca, line 16: It is true that `quarkus-quickstarts repo `main` branch uses final artifact but this is because in that repo `main` is equivalent to kogito-examples `stable` while the development is done in the `development` branch (kogito-examples `main`).
See https://github.com/quarkusio/quarkus-quickstarts/blob/development/getting-started/pom.xml#L10-L13

Btw I think the main issue is that we need to use by default `kogito-quarkus-bom` so that we can use an unreleased version of kogito with quarkus so my suggestion is to replace with
```suggestion
    <quarkus-plugin.version>${version.io.quarkus}</quarkus-plugin.version>
    <quarkus.platform.artifact-id>kogito-quarkus-bom</quarkus.platform.artifact-id>
    <quarkus.platform.group-id>org.kie.kogito</quarkus.platform.group-id>
    <quarkus.platform.version>${project.version}</quarkus.platform.version>
```

In this way the replacement command will be a bit "uglier" but we can preserve existing behaviour 🤔

README.md
```diff
@@ -4,9 +4,19 @@ This module contains a number of examples that you can take a look at and try ou
...
 - **[You can also check all versions by looking at releases.](https://github.com/kiegroup/kogito-examples/releases/latest)**
 
+## Use alternative Quarkus platforms
+
+The Quarkus quickstarts by default currently use the Quarkus core BOM.
+
+If you want to use an alternative BOM when building the Quarkus quickstarts you can override the `quarkus.platform.*` properties. The following example shows how to set `quarkus.platform.artifact-id` to use the quarkus-universe-bom.
+
+```
+mvn -Dquarkus.platform.artifact-id=quarkus-universe-bom clean install
```
danielezonca, line 17: As mentioned in our case the command will be problably
```suggestion
mvn -Dquarkus.platform.group-id=io.quarkus -Dquarkus.platform.artifact-id=quarkus-universe-bom -Dquarkus.platform.version=2.1.0.Final -Dquarkus-plugin.version=2.1.0.Final clean install
```

danielezonca (commented):
@evacchi @ricardozanini @cristianonicolai 
Please have a look to my comments and the PR, I don't think we can apply directly the quarkus-universe-bom but we need to use `kogito-quarkus-bom` with snapshot version
Wdyt?

danielezonca:
@AntStephenson 
I did the review and added two comments, I saw your message about `${version.io.quarkus}` but this code during development is used with snapshot/custom version so I think if we want to have the explicit version we can do it only with a replace during community release pipeline so that in the release branch the value is explicit while here still dynamic. Wdyt?

decisiontable-quarkus-example/pom.xml
evacchi: the optimal solution would be to have our build produce a custom `999-SNAPSHOT` quarkus platform bom and use that in development. I wonder if using `kogito-quarkus-bom` is good enough for development. /cc'ing @aloubyansky

README.md
aloubyansky: `quarkus.platform.version` is now managing both the plugin and the BOM(s), so `quarkus-plugin.version` is redundant here.

decisiontable-quarkus-example/pom.xml
aloubyansky: Are these the examples that users would be consulting and possibly copy&pasting into their projects? If so, we should be looking for a way to use the project model and BOMs we are officially promoting. Which means the official Quarkus platform BOMs. Where the universe BOM is deprecated now in favor of the `io.quarkus.platform:quarkus-bom` and `io.quarkus.platform:quarkus-kogito-bom` in your case. Given that the groupId and the version are expressed as properties, it should be possible to use different BOMs in the CI. It should also be possible to generate just these two platform BOMs as part of the CI, if necessary.

evacchi: vaguely related: https://github.com/kiegroup/kogito-runtimes/pull/1506 here we may need to do the same for testing gradle support

ricardozanini: Agreed. We should not have the quarkus version hard-coded in these POMs...

antncl: @evacchi @ricardozanini so what do you think the solution is here? should we use `io.quarkus.platform:quarkus-kogito-bom` and what about the version? originally my proposal had static versions in because of PME on the prod side but I'm happy to adjust it should we agree to have static versions on a prod branch.

ricardozanini: @evacchi `kogito-quarkus-bom` it's meant to be used **only internally**. It's our descriptor for internal [Kogito Quarkus modules](https://github.com/kiegroup/kogito-runtimes/tree/main/quarkus) only. I have a [JIRA](https://issues.redhat.com/browse/KOGITO-5691) opened to remove all the references of this BOM from our examples and ITs.

@AntStephenson, can you shed light on why we need these static versions? It's super counter-intuitive to me adding hard-coded versions to our POMs. Even with an automated process to update them in every release, it's laborious to get it done and maintain. Maybe we can come up with another solution. The common parent POM won't work?

ricardozanini: Just one last thing. +1 to remove `kogito-quarkus-bom`. We just have to come up with a solution to not have to update the Quarkus version every time.

antncl: @ricardozanini Originally the main motivation was because PME (which we use on the prod side for replacing versions with productized redhat versions and such) would not replace `${version.io.quarkus}` with a static version. But also afaik it's also possible that the platform version would not always be the same as the Quarkus version (at least on product).

ricardozanini: I understand. So if this is the only option, I would say to tie this merge with the automation in the upstream CI so that the update won't slip between our releases.

antncl: @ricardozanini @evacchi  hi, just want to confirm my intent - please speak up if you see any issue with this; 

* Update default to reference `io.quarkus.platform:quarkus-kogito-bom` as proposed by Alexey and use `${version.io.quarkus}` instead of the static versions (on main). 
* Remove the `quarkus-plugin.version` property and use the `quarkus-platform.version` for the maven plugin.
* We need automation to update to static versions on product branches.

I already spoke with Tristan before I went on PTO and he pointed me in the direction of where I can add/enhance tooling for changing the version on prod branches.

mareknovotny:
i would ask if we can merge this for next week to be in our community release.

cristianonicolai:
@AntStephenson seems like some new modules that were merged in between do not follow the new approach to define the Quarkus dependency, can you please review as that is breaking the build, see https://github.com/kiegroup/kogito-runtimes/pull/1545/checks?check_run_id=3402429180

antncl:
@cristianonicolai Hi, raised https://github.com/kiegroup/kogito-examples/pull/857 please take a look

PR #840, opened 2021-08-06
KOGITO-5667 Split Quarkus and Spring Boot specific test-utils in diff…

cristianonicolai:
…erent modules

Assembly:
- https://github.com/kiegroup/kogito-examples/pull/840
- https://github.com/kiegroup/kogito-apps/pull/976
- https://github.com/kiegroup/kogito-runtimes/pull/1510

### DIFF
--- a/dmn-event-driven-quarkus/pom.xml
+++ b/dmn-event-driven-quarkus/pom.xml
@@ -57,7 +57,7 @@
     </dependency>
     <dependency>
       <groupId>org.kie.kogito</groupId>
-      <artifactId>kogito-test-utils</artifactId>
+      <artifactId>kogito-quarkus-test-utils</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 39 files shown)


radtriste (approved):

ricardozanini (commented):
I think it's worth adding these new dependencies to Spring Boot and Quarkus BOM dependencies management section with the test scope as well. Or even as a direct dependency with test scope. Won't hurt the final user jar. wdyt?

danielezonca:
> I think it's worth adding these new dependencies to Spring Boot and Quarkus BOM dependencies management section with the test scope as well. Or even as a direct dependency with test scope. Won't hurt the final user jar. wdyt?

My suggestion is to avoid defining the scope at bom/dependencyManagement level because it can lead to strange behaviours (i.e. [link](https://stackoverflow.com/a/20750041)). I had a similar issue in some previous PR with a dependency declared with `test` scope. My suggestion is to not worry about the scope, the dependency is clearly mentioning `test` both in the name of the artifact and in the package and if a user wants to extend it in some way we should not prevent it.
At least this is my PoV :)

ricardozanini:
jenkins rerun examples tests

ricardozanini:
jenkins rerun examples tests

PR #902, opened 2021-09-27
KOGITO-5959 Public API (Incubation) Examples

evacchi:
https://issues.redhat.com/browse/KOGITO-5959

Follow up to:

- kiegroup/kogito-runtimes#1547 
- kiegroup/kogito-runtimes#1549 
- kiegroup/kogito-runtimes#1550 
- kiegroup/kogito-runtimes#1551


gitgabrio (approved):
LGTM, thx!!!

ricardozanini (approved):
Nice work man! Many thanks!

dmn-incubation-api-quarkus/src/test/java/org/acme/GreetingResourceTest.java
```diff
@@ -0,0 +1,53 @@
...
+        given()
+                .contentType(ContentType.JSON)
+                .accept(ContentType.JSON)
+                .body(body)
+                .post("/hello")
+                .then()
+                .statusCode(200)
+                .body("'Should the driver be suspended?'", is("No"));
+
+    }
```
tarilabs: ok

tarilabs (commented):
The DMN example works for me, thanks!

I was wondering if you considered opting for a test that instead of replicating the REST codegen using the new API, would have shown instead how to _directly_ use the incubation API from a user perspective.

I am fine with either choice, I just hoped this contributed something to eventually consider.

evacchi:
@tarilabs good point, I may project a specific value from the payload.. the idea for this example was to keep it really simple; there should be examples to show POJO mapping as well. I am not 100% sure whether there should be OTHER examples, or each example show a different usage (e.g. DMN + POJO with custom mapping, PMML + basic mapping. etc etc)...

evacchi:
@tarilabs I have created a JIRA to extend these examples; maybe just add another REST endpoint with customized behavior, e.g. use a POJO (gets openapi for free) project a value through the POJO and return it (i.e. use a field), use records instead of POJOs etc...

https://issues.redhat.com/browse/KOGITO-5971

evacchi:
jenkins please retest this

dmn-incubation-api-quarkus/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.kie.kogito.examples</groupId>
+    <artifactId>kogito-examples</artifactId>
+    <version>2.0.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>dmn-incubation-api-quarkus</artifactId>
+  <name>Kogito Example :: DMN Incubation API With Quarkus</name>
+  <properties>
+    <quarkus-plugin.version>${version.io.quarkus}</quarkus-plugin.version>
```
radtriste: in the rest of the examples, the version is harcoded. Could we keep that to have some consistence ?

dmn-incubation-api-quarkus/mvnw
```diff
@@ -0,0 +1,310 @@
+#!/bin/sh
```
danielezonca: In general all Kogito examples don't have `mvnw` for each module so if we want to add it we should add it for all of them.
Wdyt?
If you prefer to have it for each module can you please create a ticket to update other examples?

dmn-incubation-api-quarkus/src/main/java/org/acme/GreetingResource.java
```diff
@@ -0,0 +1,53 @@
...
+import javax.ws.rs.core.MediaType;
+
+import org.kie.kogito.incubation.application.AppRoot;
+import org.kie.kogito.incubation.common.DataContext;
+import org.kie.kogito.incubation.common.MapDataContext;
+import org.kie.kogito.incubation.decisions.DecisionIds;
+import org.kie.kogito.incubation.decisions.services.DecisionService;
+
+@Path("/hello")
+public class GreetingResource {
```
danielezonca: It is not really necessary but I would prefer to use class names and paths a bit more "specific".
What about something like?
```suggestion
@Path("/custom-traffic-violation")
public class CustomTrafficViolationResource {
```

Similar comment for each example

dmn-incubation-api-quarkus/src/main/java/org/acme/GreetingResource.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Inject
+    AppRoot appRoot;
+    @Inject
+    DecisionService svc;
+
+    @POST
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    public DataContext hello(Map<String, Object> payload) {
```
danielezonca: Same here
```suggestion
    public DataContext myCustomLogic(Map<String, Object> payload) {
```

dmn-incubation-api-quarkus/mvnw
evacchi: no that's a leftover, will remove

dmn-incubation-api-quarkus/src/main/java/org/acme/GreetingResource.java
evacchi: ok

evacchi:
jenkins run examples tests

dmn-incubation-api-quarkus/src/test/java/org/acme/GreetingResourceTest.java
evacchi: will do follow up PR with another example, e.g. a console-based application as discussed in https://issues.redhat.com/browse/KOGITO-5971

tiagodolphine (commented):
LGTM, just a comment that maybe it is worth adding some explanation of the incubation API, and the idea of the examples like using the API to expose and endpoint using this API to interact with the runtimes and skip code generation for it, something like this.

dmn-incubation-api-quarkus/.gitignore
```diff
@@ -0,0 +1,39 @@
+#Maven
```
cristianonicolai: @evacchi this file is probably not needed as we already have one at root level

dmn-incubation-api-quarkus/src/main/resources/Traffic Violation.dmn
```diff
@@ -0,0 +1,230 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
cristianonicolai: @evacchi can we remove the space in the file name?

pmml-incubation-api-quarkus/.gitignore
```diff
@@ -0,0 +1,39 @@
+#Maven
```
cristianonicolai: same here

process-incubation-api-quarkus/.gitignore
```diff
@@ -0,0 +1,39 @@
+#Maven
```
cristianonicolai: :)

process-incubation-api-quarkus/src/main/java/org/acme/CustomRestProcess.java
```diff
@@ -0,0 +1,50 @@
...
+import org.kie.kogito.incubation.processes.ProcessIds;
+import org.kie.kogito.incubation.processes.services.StraightThroughProcessService;
+
+@Path("/custom-rest-process")
+public class CustomRestProcess {
+
+    @Inject
+    AppRoot appRoot;
+    @Inject
+    StraightThroughProcessService svc;
```
cristianonicolai, line 39: @evacchi this is probably something I overlooked from the initial PR and happy to merge as is but why differentiate an STP vs regular process? That feels more like a design/model responsibility than at the API level.

evacchi: as the first supported version only supported STP and the API would not be ready to support all of `abort()`, `workItems()` `userTask()` etc we initially decided to make it clear in the name of the service, but since this is "incubation" nothing prevent us to drop it entirely and change the name

dmn-incubation-api-quarkus/src/main/resources/Traffic Violation.dmn
evacchi: we can, but this is a file that it's everywhere in all other examples :)

cristianonicolai: yeah, removed from a few places already :P

process-incubation-api-quarkus/src/main/java/org/acme/CustomRestProcess.java
cristianonicolai: makes sense, lets review it as we go, but I think a unified API would make more sense, then these methods you mentioned would have a no-op/throw exception behavour.

cristianonicolai (approved):
@evacchi great work!

process-incubation-api-quarkus/src/main/java/org/acme/CustomRestProcess.java
evacchi: we can already iterate over this in the next sprint 👍

radtriste (approved):

danielezonca (approved):

tiagodolphine (approved):

mariofusco (approved):

PR #916, opened 2021-10-06
Fix Quarkus version

cristianonicolai:


process-error-handling/pom.xml
```diff
@@ -10,10 +10,10 @@
   <name>Kogito Example :: Process Scripts With Quarkus</name>
   <description>Kogito scripts invocation - Quarkus</description>
   <properties>
-    <quarkus-plugin.version>2.1.0.Final</quarkus-plugin.version>
+    <quarkus-plugin.version>2.3.0.Final</quarkus-plugin.version>
```
radtriste, line 13: it was not updated because the module is not set in the default profile in https://github.com/kiegroup/kogito-examples/blob/main/pom.xml ... Can you please add it ?

radtriste (changes requested):

process-error-handling/pom.xml
cristianonicolai: @radtriste done

radtriste (approved):

ricardozanini (approved):
What's the cause of these updates? It's a similar problem that happened with the SW example? Not being in the `default` profile?

radtriste:
yep ... :(

cristianonicolai:
Jenkins rerun examples tests

PR #928, opened 2021-10-08
[KOGITO-6052] Protecting againts exception in abort

fjtirado:
Abort (which is added just for the persistence case) might throw an exception if the process is already removed (that's fine and can be ignored in the result of the test) 
Depends on https://github.com/kiegroup/kogito-runtimes/pull/1649


fjtirado:
jenkins retest this

fjtirado:
jenkins rerun examples tests

PR #932, opened 2021-10-13
KOGITO-5817 Disable REST codegen if resteasy is not present

evacchi:
Throws an Exception at build-time, *unless* either RESTEasy is present.

- https://github.com/kiegroup/kogito-runtimes/pull/1643
- https://github.com/kiegroup/kogito-apps/pull/1060
- https://github.com/kiegroup/kogito-examples/pull/932

http://issues.redhat.com/browse/KOGITO-5817

### DIFF
--- a/dmn-event-driven-quarkus/pom.xml
+++ b/dmn-event-driven-quarkus/pom.xml
@@ -41,6 +41,19 @@
 
     <dependency>
       <groupId>io.quarkus</groupId>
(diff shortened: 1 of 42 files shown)


danielezonca (approved):
I think it is useful to add a note in the release note of Kogito 1.13 because now RestEasy dependency must be provided 🤔 
Wdyt?

radtriste (approved):

PR #941, opened 2021-10-20
KOGITO-5841  Public API: Allow extra-metadata in DataContext

evacchi:
see parent https://github.com/kiegroup/kogito-runtimes/pull/1663


tarilabs (approved):
This will impact kogito-examples, not the Quarkus IT inside -runtimes
👍

PR #946, opened 2021-10-22
KOGITO-6175 Fixed docker networking with Macos

tfriman:
~There is no JIRA ticket for this but this does not work with Docker Desktop Version 4.1.1.~

[Jira ticket](https://issues.redhat.com/browse/KOGITO-6175)

Docker seems not to add /etc/hosts 127.0.0.1 for host.docker.internal but kubernetes.docker.internal.

### DIFF
--- a/kogito-travel-agency/extended/docker-compose/startServices.sh
+++ b/kogito-travel-agency/extended/docker-compose/startServices.sh
@@ -32,7 +32,7 @@ echo "Kogito Image version: ${KOGITO_VERSION}"
 echo "KOGITO_VERSION=${KOGITO_VERSION}" > ".env"
 
 if [ "$(uname)" == "Darwin" ]; then
-   echo "DOCKER_GATEWAY_HOST=${host.docker.internal}" >> ".env"
+   echo "DOCKER_GATEWAY_HOST=kubernetes.docker.internal" >> ".env"
 elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
    echo "DOCKER_GATEWAY_HOST=172.17.0.1" >> ".env"
 fi
--- a/process-usertasks-quarkus-with-console/docker-compose/startServices.sh
+++ b/process-usertasks-quarkus-with-console/docker-compose/startServices.sh
@@ -17,7 +17,7 @@ echo "Kogito Image version: ${KOGITO_VERSION}"
 echo "KOGITO_VERSION=${KOGITO_VERSION}" > ".env"
 
 if [ "$(uname)" == "Darwin" ]; then
-   echo "DOCKER_GATEWAY_HOST=${host.docker.internal}" >> ".env"
+   echo "DOCKER_GATEWAY_HOST=kubernetes.docker.internal" >> ".env"
 elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
    echo "DOCKER_GATEWAY_HOST=172.17.0.1" >> ".env"
 fi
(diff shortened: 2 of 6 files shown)


nmirasch:
@tfriman Thanks a lot! It seems this is what is needed.  I have a question regarding using kubernetes.docker.internal.  Is it required to 'enable kubernetes' or it is something that comes with the the docker desktop installation in Mac?. It is just a matter to add this information in the examples installation instructions.)

tfriman:
@nmirasch hi, can't be sure but I don't think I've ever enabled k8s nor experimental features. I tried searching if Docker has changed this but only found these: https://github.com/docker/for-mac/issues/4442 and https://github.com/docker/for-mac/issues/4267 If needed, I can re-install Docker Desktop (I have 4.1.1 69879, latest at time being) and check how /etc/hosts changes?

nmirasch:
@tfriman Thanks for the feedback. I read about that host.docker.internal property that would do the common IP to provide the visibility between containers but it seems it's not always there. If you tried with  'kubernetes.docker.internal' and it works, I would go ahead with the change, but I just wanted to document the needs to make it run.  
If you can do that test about the  /etc/hosts changes with the re-installation would help a lot in order to know what is exactly needed to make the containers work together. That would be really helpful!

tfriman:
@nmirasch hi, uninstalled Docker Desktop, cleaned /etc/hosts and re-installed the latest Docker Desktop 4.1.1 (69879) to my MacOS 11.6 (Big Sur) and this got added to the /etc/hosts:


&#35; Added by Docker Desktop
&#35; To allow the same kube context to work on the host and the container:
127.0.0.1 kubernetes.docker.internal
&#35; End of section


So it seems that with default settings you get the kubernetes.docker.internal only. Sadly I was not able to find any official statement about host.docker.internal deprecation so there is a chance this is not permanent though.

nmirasch:
@tfriman Thank you very much! It seems we need to rely on kubernetes.docker.internal instead. I've created this jira to track this https://issues.redhat.com/browse/KOGITO-6175. I guess the only thing missing to add the change is to update the instructions related adding that kubernetes.docker.internal reference even we can maintain as a note that  host.docker.internal for backward compatibility. WDYT?
Thanks again for all your effort and useful testing to see the examples working with the current Docker Desktop in MAC!

tfriman:
jenkins retest this

tfriman:
@nmirasch sorry missed this one. I agree, IMHO it is wise to leave that "old" host.docker.internal somewhere in comments. 
BTW, another way to fix this issue is to add instructions how to add this host.docker.internal to /etc/hosts but for time being I would change the scripts to refer directly to kubernetes.docker.internal so there would be no need to modify /etc/hosts.

tfriman:
@nmirasch got a response back from Docker support:

---
Docker Desktop doesn't add host.docker.internal to /etc/hosts. The DNS name is handled internally by a DNS proxy.

If it's useful to have an `/etc/hosts` entry on the host then I recommend:

- describe your use-case on a roadmap issue: https://github.com/docker/roadmap
- in the meantime add the entry manually
---

So seems kubernetes.docker.internal it is then!

rgmz:
I encountered a related, but not specifically mentioned, issue while testing the `process-usertasks-quarkus-with-console` example. I think `host.docker.internal` is meant to be a literal value and not an environment variable:
```sh
$ ./startServices.sh
Script requires your Kogito Quickstart to be compiled
Project version: 1.13.0.Final
Kogito Image version: 1.13.0
./startServices.sh: line 22: DOCKER_GATEWAY_HOST=${host.docker.internal}: bad substitution
```

However, `host.docker.internal` doesn't actually work for the reasons mentioned above. There's an entry for `kubenetes.docker.internal` in /etc/hosts, but not `host.docker.internal`.

Manually setting `DOCKER_GATEWAY_HOST=127.0.0.1` appears to work. I also added `127.0.0.1` as a host for the redirectUris in addition to `localhost`:
https://github.com/kiegroup/kogito-examples/blob/638d9f5ab38eb937c1a43e58e7440a6d29741bc0/process-usertasks-quarkus-with-console/docker-compose/keycloak/kogito-realm.json#L556-L558

Environment:
* macOS `11.6`
* Docker Desktop `3.6.0`
* Docker Engine `20.10.8`.

PR #977, opened 2021-11-12
adding Kogito Runtime Tools to the process-decisions-quarkus example

rafaeltuelho:
This PR extends the `process-decisions-quarkus` example adding a new maven profile (`runtime-tools`) enabling the Kogito Runtime Tools Quarkus Extension which provides Management Console in the DevUI. It also includes a docker-compose file to start the Kogito Data Index Service (with Kafka and ISPN) to support the Management Console Process Instance visualization.


danielezonca:
Ok to test

process-decisions-quarkus/README.md
```diff
@@ -181,9 +183,9 @@ Given data:
 {
     "driverId": "12345",
     "violation":{
-        "Type":"speed",
-        "Speed Limit": 100,
-        "Actual Speed":140
+        "type":"speed",
+        "speedLimit": 100,
+        "actualSpeed":140
```
tiagodolphine, line 188: I think it is better to wait for this issue to be fixed and keep the values defined in the JsonProperty. One idea of this example is to show the users how to customize attribute names and make them compatible with DMN data types.

process-decisions-quarkus/src/main/resources/application.properties
```diff
@@ -7,4 +7,41 @@ quarkus.swagger-ui.always-include=true
...
+
+org.kie.kogito.runtime.tools.quarkus.extension.runtime.dataindex.DataIndexClient/mp-rest/url=http://localhost:8180
+# Mocked users and groups for the task inbox screen
+quarkus.kogito-runtime-tools.users=jdoe,admin,user
+quarkus.kogito-runtime-tools.users.jdoe.groups=admin
+quarkus.kogito-runtime-tools.users.admin.groups=admin
+quarkus.kogito-runtime-tools.users.user.groups=user
+
+quarkus.test.continuous-testing=disabled
+quarkus.devservices.enabled=true
```
tiagodolphine: It is better to define a quarkus profile and set this profile as an env variable in the pom.xml when executing the maven profile runtime-tools you introduced, in this way, if users just execute the default example, no messaging, kafka etc will be needed in the application.properties file

tiagodolphine (commented):
@rafaeltuelho nice work, I just left some comments.
Thanks

process-decisions-quarkus/src/main/resources/application.properties
rafaeltuelho: Sure thing! That makes sense. I'll do that.

process-decisions-quarkus/README.md
rafaeltuelho: Sounds good to me.

process-decisions-quarkus/src/main/resources/application.properties
rafaeltuelho: @tiagodolphine I think I should use Quarkus default `dev` profile for these messaging and runtime-tools properties. Because the idea of the new Kogito Runtime Tools ext is to bring kogito services to Quarks DevUI. Thus when starting the app with 
```
mvn quarkus:dev -Pruntime-tools
```

will pick up these additional messaging props...
What do you think?

rafaeltuelho: Please, ignore my last comment. I decided to put these additional config props into a separate file `application-runtime-tools.properties`. This way when the user execute the `runtime-tools` maven profile  it will automatically activate a custom quarkus  profile named `runtime-tools`

tiagodolphine: great I like this idea to have a different `application.properties` for the new profile.

So let's wait for the [PR](https://github.com/kiegroup/kogito-apps/pull/1120).

domhanak:
Closing as stale

PR #1065, opened 2022-01-05
Update Notification Bot to not clip the issue link

ricardozanini:
@tarilabs @danielezonca sorry I had to open it again.

@danielezonca I've incorporated your suggestions, but I think adding the repo in the title could truncate the issue title as well, so I added it to the body instead. wdyt?


tarilabs (approved):
Imho having just the issue title was enough for the Zulip thread, but ok with me

danielezonca (approved):
Perfect :)
Please do the same also for [`kogito-runtimes`](https://github.com/kiegroup/kogito-runtimes/blob/main/.github/workflows/issues.yml) and [`kogito-apps`](https://github.com/kiegroup/kogito-apps/blob/main/.github/workflows/issues.yml)

PR #1070, opened 2022-01-07
KOGITO-6507 Removed KafkaTestClient usage in compile scope from sw-callback-quarkus

hbelmiro:
JIRA: https://issues.redhat.com/browse/KOGITO-6510

org.kie.kogito.test.quarkus.kafka.KafkaTestClient should be used only in test scope, but serverless-workflow-callback-quarkus was using it in compile scope. MicroProfile Reactive Messaging should be used rather than KafkaTestClient.

### DIFF
--- a/kogito-quarkus-examples/serverless-workflow-callback-quarkus/pom.xml
+++ b/kogito-quarkus-examples/serverless-workflow-callback-quarkus/pom.xml
@@ -34,10 +34,6 @@
     </dependency>
     <dependency>
       <groupId>org.kie.kogito</groupId>
-      <artifactId>kogito-quarkus-test-utils</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.kie.kogito</groupId>
       <artifactId>kogito-addons-quarkus-messaging</artifactId>
     </dependency>
     <dependency>
@@ -65,6 +61,11 @@
     <dependency>
       <groupId>org.awaitility</groupId>
       <artifactId>awaitility</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.kie.kogito</groupId>
+      <artifactId>kogito-quarkus-test-utils</artifactId>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/kogito-quarkus-examples/serverless-workflow-callback-quarkus/src/main/java/org/kie/kogito/examples/PublishService.java
+++ b/kogito-quarkus-examples/serverless-workflow-callback-quarkus/src/main/java/org/kie/kogito/examples/PublishService.java
@@ -22,15 +22,12 @@ import java.util.Map;
 import java.util.UUID;
 
 import javax.annotation.PostConstruct;
-import javax.annotation.PreDestroy;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
-import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.eclipse.microprofile.reactive.messaging.Channel;
+import org.eclipse.microprofile.reactive.messaging.Emitter;
 import org.kie.kogito.internal.process.runtime.KogitoProcessContext;
-import org.kie.kogito.process.Processes;
-import org.kie.kogito.test.quarkus.kafka.KafkaTestClient;
-import org.kie.kogito.testcontainers.quarkus.KafkaQuarkusTestResource;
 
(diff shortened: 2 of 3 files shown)


ricardozanini (approved):

kogito-quarkus-examples/serverless-workflow-callback-quarkus/src/main/java/org/kie/kogito/examples/PublishService.java
```diff
@@ -43,28 +40,19 @@
 public class PublishService {
 
     @Inject
-    Processes processes;
+    @Channel("outgoing-move")
```
fjtirado, line 43: you can just use "move" as channel name

kogito-quarkus-examples/serverless-workflow-callback-quarkus/src/main/resources/application.properties
```diff
@@ -3,6 +3,10 @@ kafka.bootstrap.servers=localhost:9092
 mp.messaging.incoming.move.connector=smallrye-kafka
 mp.messaging.incoming.move.value.deserializer=org.apache.kafka.common.serialization.StringDeserializer
 
+mp.messaging.outgoing.outgoing-move.connector=smallrye-kafka
+mp.messaging.outgoing.outgoing-move.value.deserializer=org.apache.kafka.common.serialization.StringDeserializer
+mp.messaging.outgoing.outgoing-move.topic=move
```
fjtirado, line 8: you can remove tha one (by default the topic name is equal to channel name) and use move as channel name

kogito-quarkus-examples/serverless-workflow-callback-quarkus/pom.xml
```diff
@@ -67,6 +63,11 @@
       <artifactId>awaitility</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
fjtirado, line 66: This kafka cliente library was used just for publishing the event, so once is not longer used, you can completely remote it, not put in test scope

fjtirado (commented):
@hbelmiro Nice change, just a couple of comments, you can remove the dependency completely and use "move" as channel name to save one property

kogito-quarkus-examples/serverless-workflow-callback-quarkus/src/main/java/org/kie/kogito/examples/PublishService.java
hbelmiro: I think I can't because AFAIK we can't use the same channel name for outgoing and incoming. So, I named the outgoing channel as "outgoing-move" and let the incoming channel with the name it already had ("move").

kogito-quarkus-examples/serverless-workflow-callback-quarkus/src/main/resources/application.properties
hbelmiro: Because of what I explained above, the channel name is actually "outgoing-move". So, we have two channels pointing to the same topic:
| Channel | Type | Topic |
| ----------- | ------- | ------- |
| move | incoming | move |
| outgoing-move | outgoing | move |

kogito-quarkus-examples/serverless-workflow-callback-quarkus/pom.xml
hbelmiro: It's still used in test scope. It provides the classes `KafkaQuarkusTestResource` and `PostgreSqlQuarkusTestResource` for `CallbackRestIT`.

kogito-quarkus-examples/serverless-workflow-callback-quarkus/src/main/java/org/kie/kogito/examples/PublishService.java
fjtirado: You are right
SRMSG00073: Invalid configuration, the following channel names cannot be used for both incoming and outgoing:

fjtirado (approved):

kogito-quarkus-examples/serverless-workflow-callback-quarkus/src/main/resources/application.properties
fjtirado: yes, I believe (wrongly) that the same channel name can be used both for incoming and outgoing, but thats not the case

hbelmiro:
Thanks for the review @fjtirado

PR #1101, opened 2022-01-24
[1.11.x] Review prod modules

cristianonicolai:


jstastny-cz (approved):
The list of modules included here are in sync with 1.5.x with following additions:
rules-legacy-quarkus-example
rules-legacy-springboot-example

Which IMO is expected.

PR #1052, opened 2021-12-20
KOGITO-6090 detach examples from kogito-build-parent

jstastny-cz:
https://issues.redhat.com/browse/KOGITO-6090

Ensemble with kiegroup/kogito-runtimes#1811

Changing Quarkus-related examples so that:
* BOM import and management is done by modules themselves (so far kogito-build-parent brought in kogito-bom import transitively).
* They rely on io.quarkus:quarkus-bom and org.kie.kogito:kogito-bom.
* The GAV configuration for both is parameterized using properties.
* Which allows easily execute kogito-examples with Quarkus Platform BOMs.

<details>
<summary>
How to replicate CI configuration locally?
</summary>

Build Chain tool does "simple" maven build(s), the builds are just Maven commands, but because the repositories relates and depends on each other and any change in API or class method could affect several of those repositories there is a need to use [build-chain tool](https://github.com/kiegroup/github-action-build-chain) to handle cross repository builds and be sure that we always use latest version of the code for each repository.
 
[build-chain tool](https://github.com/kiegroup/github-action-build-chain) is a build tool which can be used on command line locally or in Github Actions workflow(s), in case you need to change multiple repositories and send multiple dependent pull requests related with a change you can easily reproduce the same build by executing it on Github hosted environment or locally in your development environment. See [local execution](https://github.com/kiegroup/github-action-build-chain#local-execution) details to get more information about it.
</details>

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>Pull Request</b>  
  Please add comment: <b>Jenkins retest this</b>
 
* <b>Quarkus LTS checks</b>  
  Please add comment: <b>Jenkins run LTS</b>

* <b>Native checks</b>  
  Please add comment: <b>Jenkins run native</b>

</details>

PR #1113, opened 2022-02-02
[1.11.x] KOGITO-6090 detach examples from kogito-build-parent

jstastny-cz:
[KOGITO-6090](https://issues.redhat.com/browse/KOGITO-6090)
1.11.x variant of #1052 (not a cherry-pick)

<details>
<summary>
How to replicate CI configuration locally?
</summary>

Build Chain tool does "simple" maven build(s), the builds are just Maven commands, but because the repositories relates and depends on each other and any change in API or class method could affect several of those repositories there is a need to use [build-chain tool](https://github.com/kiegroup/github-action-build-chain) to handle cross repository builds and be sure that we always use latest version of the code for each repository.
 
[build-chain tool](https://github.com/kiegroup/github-action-build-chain) is a build tool which can be used on command line locally or in Github Actions workflow(s), in case you need to change multiple repositories and send multiple dependent pull requests related with a change you can easily reproduce the same build by executing it on Github hosted environment or locally in your development environment. See [local execution](https://github.com/kiegroup/github-action-build-chain#local-execution) details to get more information about it.
</details>

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>Pull Request</b>  
  Please add comment: <b>Jenkins retest this</b>
 
* <b>Quarkus LTS checks</b>  
  Please add comment: <b>Jenkins run LTS</b>

* <b>Native checks</b>  
  Please add comment: <b>Jenkins run native</b>

</details>

### DIFF
--- a/decisiontable-quarkus-example/pom.xml
+++ b/decisiontable-quarkus-example/pom.xml
@@ -14,6 +14,9 @@
     <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
     <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
     <quarkus.platform.version>2.2.3.Final</quarkus.platform.version>
+    <kogito.bom.group-id>org.kie.kogito</kogito.bom.group-id>
+    <kogito.bom.artifact-id>kogito-bom</kogito.bom.artifact-id>
+    <kogito.bom.version>${project.version}</kogito.bom.version>
   </properties>
   <dependencyManagement>
     <dependencies>
@@ -21,6 +24,13 @@
         <groupId>${quarkus.platform.group-id}</groupId>
         <artifactId>${quarkus.platform.artifact-id}</artifactId>
         <version>${quarkus.platform.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+      <dependency>
+        <groupId>${kogito.bom.group-id}</groupId>
+        <artifactId>${kogito.bom.artifact-id}</artifactId>
+        <version>${kogito.bom.version}</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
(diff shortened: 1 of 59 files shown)


antncl (approved):

jstastny-cz:
Jenkins retest this

jstastny-cz:
Jenkins retest this.

jstastny-cz:
jenkins retest this

jstastny-cz:
jenkins retest this

cristianonicolai (approved):

decisiontable-quarkus-example/pom.xml
```diff
@@ -14,6 +14,9 @@
     <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
     <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
     <quarkus.platform.version>2.2.3.Final</quarkus.platform.version>
```
MarianMacik, line 16: Why not to use the `version.io.quarkus` coming from the `kogito-dependencies-bom`? It would be much easier to maintain.

onboarding-example/pom.xml
```diff
@@ -31,6 +34,13 @@
...
+      <dependency>
+        <groupId>${kogito.bom.group-id}</groupId>
+        <artifactId>${kogito.bom.artifact-id}</artifactId>
+        <version>${kogito.bom.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
     </dependencies>
   </dependencyManagement>
   <build>
```
MarianMacik, line 46: `<groupId>` of the plugin not configurable here.

MarianMacik (approved):
2 suggestions, otherwise looks good.

decisiontable-quarkus-example/pom.xml
jstastny-cz: As far as I understand it, having versions default values explicitly here is an convenience to users of examples - the configuration of a project in here should be close to what user should do on their side. (That's why we don't specify these versions only in root pom.xml, but rather in each one of the modules).
IMO kogito-dependencies-bom is not supposed to be user-facing BOM, so taking version from there might be puzzling.

MarianMacik: OK, in that case it makes sense. Though having the version at least in the root pom.xml would make maintenance much easier.

PR #1326, opened 2022-07-11
[KOGITO-7535] Test enum in grpc example

fjtirado:
Depends on https://github.com/kiegroup/kogito-runtimes/pull/2326


MarianMacik (approved):
Looks good to me.

ricardozanini (approved):
Even though we are using the examples as IT, I think the runtime could have a unit/integration test there as well. wdyt?

fjtirado:
@ricardozanini  Im going to close this one and add integration test to runtimes, keeping the examples unchanged

PR #1389, opened 2022-09-07
Remove unnecessary extension kogito-addons-quarkus-persistence-jdbc

cristianonicolai:
Need to be removed due to capability check now enabled in https://github.com/kiegroup/kogito-runtimes/pull/2442. Doesnt seems to be used anyway.

### DIFF
--- a/serverless-workflow-examples/serverless-workflow-newsletter-subscription/subscription-service/pom.xml
+++ b/serverless-workflow-examples/serverless-workflow-newsletter-subscription/subscription-service/pom.xml
@@ -173,10 +173,6 @@
       </activation>
       <dependencies>
         <dependency>
-          <groupId>org.kie.kogito</groupId>
-          <artifactId>kogito-addons-quarkus-persistence-jdbc</artifactId>
-        </dependency>
-        <dependency>
           <groupId>io.quarkus</groupId>
           <artifactId>quarkus-jdbc-postgresql</artifactId>
         </dependency>


wmedvede:
Hi @cristianonicolai , I think @ricardozanini can give more details about this particular example, but if I'm not wrong, the example is prepared intentionally to use persistence when it's executed in knative + minikube. 
Persistence configurations are set in https://github.com/kiegroup/kogito-examples/blob/stable/serverless-workflow-examples/serverless-workflow-newsletter-subscription/subscription-flow/src/main/resources/application-knative.properties

Why are we removing the persistence plugin?

wmedvede (commented):
Maybe I'm missing something, but don't understand the removal.
See my comment.

cristianonicolai:
@wmedvede subscription-flow is using the persistence as expected, the issue here is in the service application, which is not a kogito service and it is using a kogito addon. From what I can see it shouldnt be an issue to remove it, as peristence deps are still added in https://github.com/kiegroup/kogito-examples/blob/stable/serverless-workflow-examples/serverless-workflow-newsletter-subscription/subscription-service/pom.xml#L181. Unless I missing something :)

wmedvede:
> @wmedvede subscription-flow is using the persistence as expected, the issue here is in the service application, which is not a kogito service and it is using a kogito addon. From what I can see it shouldnt be an issue to remove it, as peristence deps are still added in https://github.com/kiegroup/kogito-examples/blob/stable/serverless-workflow-examples/serverless-workflow-newsletter-subscription/subscription-service/pom.xml#L181. Unless I missing something :)

Ah!, yes, I thought it was about to be removed from the project with the SW. Yes, this other project doesn't contain any SW, so it can be removed.

wmedvede (approved):

radtriste (approved):

PR #1647, opened 2023-04-20
[KOGITO-9031] Removing test from example

fjtirado:
The Readme is still valid, but the way the doc is generated implies the annotation is not longer in the rest resource, making the current test invalid, hence removing.
Merge with https://github.com/kiegroup/kogito-runtimes/pull/2945


hbelmiro (commented):
Rather than removing the test we could download the OpenAPI document and check if the tags are there.
I know it would be painful parsing the document and I'm not sure it's worth it doing this now. Maybe we can create a Jira to do it later when we have some time.
WDYT?

fjtirado:
> Rather than removing the test we could download the OpenAPI document and check if the tags are there. I know it would be painful parsing the document and I'm not sure it's worth it doing this now. Maybe we can create a Jira to do it later when we have some time. WDYT?

@hbelmiro I agree, and I fact I actually wrote a test that invoked "http://localhost:8080/q/swagger-ui/" but it does work with 404 because the swagger is not generated for quarkus test. Therefore we need to investigate it further

hbelmiro:
> > Rather than removing the test we could download the OpenAPI document and check if the tags are there. I know it would be painful parsing the document and I'm not sure it's worth it doing this now. Maybe we can create a Jira to do it later when we have some time. WDYT?
> 
> @hbelmiro I agree, and I fact I actually wrote a test that invoked "http://localhost:8080/q/swagger-ui/" but it does work with 404 because the swagger is not generated for quarkus test. Therefore we need to investigate it further

@fjtirado IIRC, there's a way to get the file without invoking swagger-ui, but I'd have to research it. I'll create a Jira.

-----

### Edit

Jira: https://issues.redhat.com/browse/KOGITO-9032

hbelmiro (approved):

PR #1666, opened 2023-05-15
[KOGITO-1477] bump maven version to 3.8.6

albfan:
match maven minimum version

required at least on https://github.com/kiegroup/kogito-examples/tree/main/kogito-quarkus-examples/process-scripts-quarkus

### DIFF
--- a/.mvn/wrapper/maven-wrapper.properties
+++ b/.mvn/wrapper/maven-wrapper.properties
@@ -1,2 +1,2 @@
-distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.8.4/apache-maven-3.8.4-bin.zip
+distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.8.6/apache-maven-3.8.6-bin.zip
 wrapperUrl=https://repo.maven.apache.org/maven2/io/takari/maven-wrapper/0.5.6/maven-wrapper-0.5.6.jar


fjtirado:
@radtriste @cristianonicolai 
Should we still use maven wrapper or is it better to remove it?

cristianonicolai:
Personally, I don't use it, but not sure if others find it useful.

radtriste:
@fjtirado wrapper are not used by the CI. Not sure if we want to follow that direction. Could be useful...
I have a long running issue to add wrapper a bit everywhere to follow the Maven version but never had time to complete it
I guess it does not hurt yet to keep it.

With the possible move to Apache, we will need to remove any binaries anyway so the jar cannot be there but the rest can (properties + java to download)

albfan:
I realize I'm using fedora 36 and that comes with maven 3.8.4. Fedora 38 (official release) has 3.8.6 so after upgrade It will be fine to use mvn directly.

I'm ok with removing this wrapper. In any case, if it is available, match minimum maven version required should be just fine, no?

fjtirado:
I vote for removing this directory so one thing less to do (remember to upgrade it to the minimum required version)
If users want to sue mvn wrapper for some reason, they can run it manually.

porcelli (approved):

baldimir:
I also vote for removing the wrapper. If we decide to remove it, we need to make sure it is replaced also in the documentation and in READMEs if it is mentioned somewhere.

porcelli:
wrapper is easily forgotten to be updated as this PR shows

domhanak:
Closing as stale, current maven we use is `3.9.6` so this is not relevant.
Please follow up here - https://github.com/apache/incubator-kie-kogito-examples/issues/1890

PR #1671, opened 2023-05-24
KOGITO-9211 - Build images needed to run IT tests without a profile

cristianonicolai:


radtriste:
I fear that this change will break some other pipelines with no docker installed there ...

radtriste:
forgot to approve XD

PR #1735, opened 2023-07-07
KOGITO-9549 - Use fixed version for docker image version

cristianonicolai:
Align image version between Jib and Docker Compose, following the same pattern as other SpringBoot examples.

### DIFF
--- a/kogito-springboot-examples/process-usertasks-springboot-with-console/docker-compose/docker-compose.yml
+++ b/kogito-springboot-examples/process-usertasks-springboot-with-console/docker-compose/docker-compose.yml
@@ -118,7 +118,7 @@ services:
 
   process-usertasks-springboot-with-console:
     container_name: process-usertasks-springboot-with-console
-    image: org.kie.kogito.examples/process-usertasks-springboot-with-console:${KOGITO_VERSION}
+    image: org.kie.kogito.examples/process-usertasks-springboot-with-console:1.0
     ports:
       - 8080:8080
     depends_on:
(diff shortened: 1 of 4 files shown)


nmirasch (approved):

AlpsDevABE:
I don't think this is the appropriate change given the README.txt

By changing ${KOGITO_VERSION} to 1.0 it won't fix the issue. 
The readme states:

> **_NOTE:_**  This step requires the project to be compiled, please consider running a ```mvn clean install -Pcontainer``` command on the project root before running the ```startServices.sh``` script for the first time or any time you modify the project. ```-Pcontainer``` will create a local docker image of the example and will be run by docker-compose with all required services.

By running this command it is going to take the verison numver from the pom file and later when it creates the docker image will tag it with 1.40.0.Final.

The docker image here is getting Kogito_Version from the startServices.sh script which runs:

PROJECT_VERSION=$(cd ../ && mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
KOGITO_VERSION=${PROJECT_VERSION%.*}

This script is going to yield: 1.40.0 without the .final

By hardcoding it to 1.0 I'm not sure what you are solving for exactly. 
The change should either be in the POM file or the startServices.sh script.

AlpsDevABE:
Just to clairfy the issue:

Docker_compose script is looking for the tag: 1.40.0 but the mvn cmd is making the docker image with the tag 1.40.0.Final

AlpsDevABE:
Oh, I'm sorry I didn't see the change in the POM.XML. 
Disregard my comments about it not working.
However wouldn't hurt to re-investigate the full flow as the startServices.sh is technically involved in the cause of the original problem.

radtriste (approved):

PR #1745, opened 2023-07-24
Adding schema/complex.json to native resources

fjtirado:
Till we found a better automatic way to do that we need to explicitly include the references resource in the native set
A more comprehesive solution will be done with https://issues.redhat.com/browse/KOGITO-9594, which will refactor the way external resources are loaded.


radtriste (approved):

radtriste:
jenkins run native

cristianonicolai (approved):

PR #1767, opened 2023-08-17
KOGITO-8400 Setup Makefile

radtriste:
https://issues.redhat.com/browse/KOGITO-8400

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,16 +19,16 @@ build-quickly:
 	$(mvn_cmd) clean install -Dquickly
 
 .PHONY: build-upstream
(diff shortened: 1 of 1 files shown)


ricardozanini (approved):

jstastny-cz (commented):
Why are some targets phony and others not?
For the phony targets - would it mean that you expect that files named `build, build-quickly, test,...` might appear in the folder and thus prevent execution of the rule?

cristianonicolai (approved):

radtriste:
> Why are some targets phony and others not? For the phony targets - would it mean that you expect that files named `build, build-quickly, test,...` might appear in the folder and thus prevent execution of the rule?

That is my bad. All targets should be PHONY so we make sure it is executed

jstastny-cz (approved):

PR #1867, opened 2024-02-07
Fix #1866 - Remove Kogito Operator CRs and citations

### LINKED ISSUE #1866: Remove references from the Kogito Operator
See https://github.com/apache/incubator-kie-kogito-operator/issues/1537

ricardozanini:
See https://github.com/apache/incubator-kie-kogito-operator/issues/1537

I've removed the Trusty Demonstration example since it relies on the Kogito Operator and pretty old binaries and libraries. To maintain it in the repo, it would have to upgrade/review the whole demo. 

The SonataFlow Github Showcase was not removed because we deprecated and there's the intention to bring it back, I just need to prioritize.

Every citation and operator's CRs were removed for good.


ricardozanini:
@lucamolteni @gitgabrio not sure who's responsible for Trusty now? Just to let you know I've removed the Trusty Demo from the repo.

gitgabrio:
@ricardozanini IINW, I'm the only one left of the "trusty" initiative, but I dealt almost only on the PMML implementation.
About Trusty (and Trusty Demo) I think there was interesting stuff in it, that we may want to retrieve when we have some time. With that in mind, I think it would be better to "disable" the Trusty Demo project (i.e. remove from reactor) instead of completely remove it. If we delete it now, it would be hard in future to remember it was there and how it was done, unless we start looking backard in commits ? WDYT ? <- @porcelli

lucamolteni:
> @lucamolteni @gitgabrio not sure who's responsible for Trusty now? Just to let you know I've removed the Trusty Demo from the repo.

@ruivieira @RobGeada ?

domhanak (approved):
Ok from serverless side.

ricardozanini:
@gitgabrio I can revert it, but honestly, it can bring more harm than good since it's super old and won't work at all. It dates back to Kogito 1.7.

gitgabrio:
> @gitgabrio I can revert it, but honestly, it can bring more harm than good since it's super old and won't work at all. It dates back to Kogito 1.7.

@ricardozanini I see your point, ok

porcelli (approved):
lgtm! thank you

wmedvede (approved):

pefernan (approved):

jstastny-cz (approved):
LGTM.

PR #1924, opened 2024-05-21
kie-issues#1123: add snapshots repository to resolve parent

jstastny-cz:
Adjustment for
* apache/incubator-kie-issues#1123

When a project relies on snapshot parent, it needs to define (at least) snapshot repositories to get the parent from. Otherwise a settings.xml repository entry would be needed to get users going.

the other case, kogito-apps have been already addressed as part of
* apache/incubator-kie-kogito-apps#2049

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -118,6 +118,18 @@
     <version.maven.artifact.plugin>3.4.1</version.maven.artifact.plugin>
   </properties>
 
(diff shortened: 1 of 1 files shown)


tiagobento (approved):

rodrigonull (approved):
LGTM

PR #1926, opened 2024-05-22
[incubator-kie-issues#1249] Fix REST call in process-rest-workitem-quarkus example

martinweiler:
Fixes https://github.com/apache/incubator-kie-issues/issues/1249

Some points regarding the changes to fix this example:
* Port needed to change from 80 to 8080 (added parameter in the WIH of the process)
* The previous version of the example used Quarkus REST Client, this would have required a config change
* With the REST client approach, the REST call was targeting `https://petstore.swagger.io/v2/user/{user}`. This call might have returned 200 for a demo user in the past, but now always returns 404. This makes it unusable for the example.
* To simplify the example, I replaced the usage of the REST Client using petstore with a simple REST service within the application.

### DIFF
--- a/kogito-quarkus-examples/process-rest-workitem-quarkus/pom.xml
+++ b/kogito-quarkus-examples/process-rest-workitem-quarkus/pom.xml
@@ -81,10 +81,6 @@
     </dependency>
     <dependency>
       <groupId>io.quarkus</groupId>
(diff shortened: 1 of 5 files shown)


elguardian (approved):
LGTM

pefernan (approved):

LightGuard:
Thanks for the fix, @martinweiler!

PR #1988, opened 2024-07-30
Updating quarkus examples pom.xml files

LightGuard:
Updating the pom files to remove properties which make it confusing for end users. This removes most properties in the pom. The properties removed are only there the change the gav for some dependencies.

### DIFF
--- a/kogito-quarkus-examples/decisiontable-quarkus-example/pom.xml
+++ b/kogito-quarkus-examples/decisiontable-quarkus-example/pom.xml
@@ -29,37 +29,26 @@
   </parent>
   <artifactId>decisiontable-quarkus-example</artifactId>
   <name>Kogito Example :: Decision Table - Quarkus</name>
-  <properties>
-    <quarkus-plugin.version>3.8.4</quarkus-plugin.version>
-    <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
-    <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
-    <quarkus.platform.version>3.8.4</quarkus.platform.version>
-    <kogito.bom.group-id>org.kie.kogito</kogito.bom.group-id>
-    <kogito.bom.artifact-id>kogito-bom</kogito.bom.artifact-id>
-    <kogito.bom.version>999-SNAPSHOT</kogito.bom.version>
-    <version.org.kie.kogito>999-SNAPSHOT</version.org.kie.kogito>
-    <version.org.drools>999-SNAPSHOT</version.org.drools>
-  </properties>
   <dependencyManagement>
     <dependencies>
       <dependency>
-        <groupId>${quarkus.platform.group-id}</groupId>
-        <artifactId>${quarkus.platform.artifact-id}</artifactId>
-        <version>${quarkus.platform.version}</version>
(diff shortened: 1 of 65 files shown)


jomarko:
I will review this tomorrow and check the relationship for the exiting PR https://github.com/apache/incubator-kie-kogito-examples/pull/1984. That, probably will be closed, need to check.

kogito-quarkus-examples/decisiontable-quarkus-example/pom.xml
```diff
@@ -29,37 +29,26 @@
...
-  </properties>
   <dependencyManagement>
     <dependencies>
       <dependency>
-        <groupId>${quarkus.platform.group-id}</groupId>
-        <artifactId>${quarkus.platform.artifact-id}</artifactId>
-        <version>${quarkus.platform.version}</version>
+        <groupId>io.quarkus</groupId>
+        <artifactId>quarkus-bom</artifactId>
+        <version>3.8.4</version>
```
rgdoliveira, line 37: isn't it better to keep versions defined as part of properties instead of moving them to the dependency block? 
I agree that some properties you removed make sense to be removed to make it simple, but maybe keep the versions in the properties block?

mareknovotny:
hard-coding of version is always not good as that requires manual changes on multiple places later and regularly.

LightGuard:
I fully understand the argument, and I thought about it. I typically do just that as well. However, when I was doing this one, I looked hard at:

1) how many places do we have this version? Answer: One place. Does it matter if it's in a version tag or a property tag? It is still one place we have to change it. 
2) these are examples for people using our software, probably for the first time. They may not be familiar with Maven (I know that sounds odd, but it is true). The easier we can make these examples to follow and use for future starting places for users, the more I think it will work out to our advantage.

kogito-quarkus-examples/decisiontable-quarkus-example/pom.xml
fjtirado: In my opinion, version (Kogito one, which is used twice in the same file and even quarkus one, which is used once, that way the two will be together) should be kept in properties block. It will be easier to change.

fjtirado:
> I fully understand the argument, and I thought about it. I typically do just that as well. However, when I was doing this one, I looked hard at:
> 
>     1. how many places do we have this version? Answer: One place. Does it matter if it's in a version tag or a property tag? It is still one place we have to change it.
> 
>     2. these are examples for people using our software, probably for the first time. They may not be familiar with Maven (I know that sounds odd, but it is true). The easier we can make these examples to follow and use for future starting places for users, the more I think it will work out to our advantage.

LightGuard:
Why was the PR closed?

fjtirado:
> Why was the PR closed?

A home accident, reopened. Apologies

ricardozanini (changes requested):
As stated in the mailing list by @gitgabrio, removing parent and version properties, even for examples, is not the best approach. The Java community has been using this approach for years, and as exemplified in the list, Spring Framework (the most popular Java framework out there) also uses it.

gitgabrio (changes requested):
Thanks @LightGuard , but as I stated elsewhere, IMO this refactoring goes exactly other way respect maintainability

LightGuard:
I'll close this one and create a new PR.

PR #1991, opened 2024-08-01
Increasing devservices timeout for keycloak

fjtirado:
This should fix [flaky test](https://ci-builds.apache.org/job/KIE/job/kogito/job/10.0.x/job/nightly/job/kogito-examples.build-and-deploy/17/) popping up

### DIFF
--- a/serverless-workflow-examples/serverless-workflow-oauth2-orchestration-quarkus/acme-financial-service/src/main/resources/application.properties
+++ b/serverless-workflow-examples/serverless-workflow-oauth2-orchestration-quarkus/acme-financial-service/src/main/resources/application.properties
@@ -19,6 +19,7 @@
 
 quarkus.http.port=8483
 quarkus.native.native-image-xmx=8g
+quarkus.devservices.timeout=2m
 
 quarkus.swagger-ui.always-include=true
 


serverless-workflow-examples/serverless-workflow-oauth2-orchestration-quarkus/acme-financial-service/src/main/resources/application.properties
```diff
@@ -19,6 +19,7 @@
 
 quarkus.http.port=8483
 quarkus.native.native-image-xmx=8g
+quarkus.devservices.timeout=2m
```
gitgabrio, line 22: Hi @fjtirado 
I see the point, but are we saying that this single test could hang two minutes before failing ?
Is it not a code-smell of some underlying issue ?
And if a similar problem arise on other tests, are we going to raise the timeout in the same way, getting to a never-ending build ?
My questions are not related to this PR in particular, for which I've not a strong opinion; but we are keeping increasing timeouts, lately, so maybe we should start looking at the issue at its root.
@baldimir any opinion on that ?

fjtirado: @gitgabrio Jenkins is slow and downloading the keycloak image and deploying it takes more than the 60s default. 
It has nothing to be with our codebase. See Apache list for more information on that. There I already complanined about our test strategy, but this one has to go in or the nightly is going to unnecesarily be flaky @pibizza

PR #1996, opened 2024-08-08
Removing artifact and group id properties

LightGuard:
Setting the affected properties to their kogito and quarkus community references.

Versions and Parent poms are left alone.


LightGuard:
How do I trigger the build here? The failures look unrelated.

LightGuard:
jenkins retest this

PR #2005, opened 2024-08-26
Include usage of function returning not json objects in sonataflow examples

fjtirado:
This illustrate the usage of funcions that returns not json objects, both for rest and jq functions


serverless-workflow-examples/serverless-workflow-temperature-conversion/multiplication-service/src/main/resources/application.properties
```diff
@@ -21,6 +21,7 @@
 # quarkus.package.type=fast-jar
 quarkus.native.native-image-xmx=8g
 quarkus.swagger-ui.always-include=true
+quarkus.devservices.enabled=false
```
fjtirado, line 24: Although unrelated, we do not need docker to run this examples

tiagobento:
Can you please improve the title and the description of this PR? Since it doesn't link to any issue, more details describing the intentions and motives here would be good. Thanks!

ricardozanini (approved):

PR #2019, opened 2024-10-04
[Fix #3694] Modifying for each example

fjtirado:
Fix https://github.com/apache/incubator-kie-kogito-runtimes/issues/3694
Updates foreach example, replacing loop variable access. Before it was accesed as property using ".", now is accessed as variable using "$"
Merge with https://github.com/apache/incubator-kie-kogito-runtimes/pull/3696


gmunozfe (approved):
Looks good to me, thx @fjtirado

ricardozanini (approved):

PR #2115, opened 2025-07-23
[10.1.x] NO_ISSUE: Update version to 10.1.0

domhanak:
Release is out, let's update our branch to point to it.

Please also merge this - https://github.com/apache/incubator-kie-kogito-examples/pull/2116

### DIFF
--- a/kogito-java-examples/dmn-embedded-mode-example/pom.xml
+++ b/kogito-java-examples/dmn-embedded-mode-example/pom.xml
@@ -24,7 +24,7 @@
     <groupId>org.kie.kogito.decisions.embedded</groupId>
     <artifactId>dmn-embedded-mode-example</artifactId>
     <name>Kogito Example :: DMN Embedded Mode</name>
-    <version>10.1.999-SNAPSHOT</version>
+    <version>10.1.0</version>
     <packaging>kjar</packaging>
 
     <properties>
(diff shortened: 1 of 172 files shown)


Rikkola:
@domhanak I am not familiar with our release methods so need to ask few things. 
1. Do we tag the release with the non-snapshot version?
2. Is there a reason why we set the branch to hard non-snapshot version?

domhanak:
@Rikkola So afaik, the current established way for examples is we just keep the branch paired to release and work with it as stable.
For 10.0.0 we fixed the version to `10.0.0` in `10.0.x` and deem the `10.0.x` as stable branch for examples.

So  I am following same pattern.
Fix 10.1.0 to released version -> `10.1.0` 
Update README to direct users to this branch for 10.1.0 examples.

1. We did not do any tagging, feel free to do so though or suggest it.

2. The reason we set it to non-snapshot is that we consider his a stable branch for example for `10.1.0`. It was set to snapshot version so that we could do some testing prior to the release.

The idea is that the examples are not published anywhere but reside in the repo to help quickstart the projects, so this seems like a easiest way currently.

Rikkola (approved):
Thanks for the info, makes sense. This way the docs have a release cycle of their own and that works out.

fjtirado (approved):

PR #2197, opened 2026-04-02
[incubator-kie-issues#2029] Dependency management simplification

### LINKED ISSUE #2029: Dependency management simplification
This issue is to track effort to properly separate the BoMs for Quarkus and Springboot in Drools.

Ideally it should be possible to have a BoM for drools/quarkus projects, a BoM for drools/Springboot projects and a BoM for pure Drools project.

As demontrated by PRs on quarkus upgrade, version of the same libraries are declared/duplicated in multiple boms/poms, making harder to maintain/keep in sync them all over the repositories.

Scope of this ticket is to:

- identify all of them
- group all of them in one single pom.xml inside the "drools" repository and have all of them inherited throughout the codebase, so that such version would be declared only in one single place
- one of the observed pattern is that the version is declared in some pom, and then the library itself is declared in some dependencyManagement tag in children pom - which some how break the concept of dependencyManagement in two different places (the pom where the version is declared as property and the one where it is used in the dependencyManagement) and lead to the duplication - ideally, we should need one single dependencyManagement declaration all over the stack
- enforce clearly separated pom hierarchies between quarkus and springboot frameworks

This comes from suggestions of @gitgabrio and discussions with @mariofusco

gitgabrio:
Fixes: https://github.com/apache/incubator-kie-issues/issues/2029

Depends on:
https://github.com/apache/incubator-kie-drools/pull/6633
https://github.com/apache/incubator-kie-kogito-runtimes/pull/4241
https://github.com/apache/incubator-kie-kogito-apps/pull/2317


kogito-springboot-examples/dmn-15-springboot-example/pom.xml
```diff
@@ -41,12 +41,12 @@
...
   <dependencyManagement>
     <dependencies>
-      <dependency>
+      <!--<dependency>
         <groupId>org.kie</groupId>
         <artifactId>kie-dmn-test-resources</artifactId>
         <version>${project.version}</version>
         <classifier>tests</classifier>
-      </dependency>
+      </dependency>-->
```
yesamer: @gitgabrio Can we remove this commented declaration?

pom.xml
```diff
@@ -386,6 +388,14 @@
             <outputTimestamp>${project.build.outputTimestamp}</outputTimestamp>
           </configuration>
         </plugin>
+        <plugin>
+          <!-- So, how to consider kogito-examples ? Shall we manage as the upstream code, enforcing rules, or as templates for final users, where such rules may not make sense ? -->
```
yesamer: @gitgabrio In my opinion, the current proposal makes sense. We should enforce this rule only within our stack; the leaf POM in the end user project should not inherit it. I think the comment can be removed.

kogito-quarkus-examples/process-instance-migration-quarkus/pom.xml
```diff
@@ -36,8 +36,8 @@
     <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
     <quarkus.platform.version>3.27.2</quarkus.platform.version>
     <kogito.bom.group-id>org.kie.kogito</kogito.bom.group-id>
-    <kogito.bom.artifact-id>kogito-bom</kogito.bom.artifact-id>
-    <kogito-apps.bom.artifact-id>kogito-apps-bom</kogito-apps.bom.artifact-id>
+    <kogito.bom.artifact-id>kogito-quarkus-bom</kogito.bom.artifact-id>
```
yesamer, line 39: @gitgabrio 
- Shouldn't this be propagated to all Quarkus based examples? Why 2 examples only?
- Isn't the quarkus BOM already imported by kogito-quarkus-bom?

kogito-springboot-examples/pom.xml
```diff
@@ -36,18 +36,9 @@
 
   <properties>
     <java.module.name>org.kie.kogito.examples.springboot</java.module.name>
+    <version.org.springframework.boot>3.5.10</version.org.springframework.boot>
```
yesamer: @gitgabrio Why this example only requires it?

kogito-quarkus-examples/process-instance-migration-quarkus/pom.xml
gitgabrio: @tiagobento FYI

kogito-springboot-examples/pom.xml
gitgabrio: @tiagobento FYI

yesamer (approved):
Thank you @gitgabrio. I have the impression that 

```
<dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>${quarkus.platform.group-id}</groupId>
        <artifactId>${quarkus.platform.artifact-id}</artifactId>
        <version>${quarkus.platform.version}</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
      <dependency>
        <groupId>${kogito.bom.group-id}</groupId>
        <artifactId>${kogito.bom.artifact-id}</artifactId>
        <version>${kogito.bom.version}</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
```

Could be removed by quarkus examples, BUT considering those example should help end user to create a new project from green field, I guess we can keep it... . I guess we should just import kogito-apps-quarkus-bom there

gitgabrio:
> Thank you @gitgabrio. I have the impression that
> 
> ```
> <dependencyManagement>
>     <dependencies>
>       <dependency>
>         <groupId>${quarkus.platform.group-id}</groupId>
>         <artifactId>${quarkus.platform.artifact-id}</artifactId>
>         <version>${quarkus.platform.version}</version>
>         <type>pom</type>
>         <scope>import</scope>
>       </dependency>
>       <dependency>
>         <groupId>${kogito.bom.group-id}</groupId>
>         <artifactId>${kogito.bom.artifact-id}</artifactId>
>         <version>${kogito.bom.version}</version>
>         <type>pom</type>
>         <scope>import</scope>
>       </dependency>
>     </dependencies>
>   </dependencyManagement>
> ```
> 
> Could be removed by quarkus examples, BUT considering those example should help end user to create a new project from green field, I guess we can keep it... . I guess we should just import kogito-apps-quarkus-bom there

Yes, it would be possible to refactor the examples a bit, but I would prefer to do that on a second step.
In that very case, there are all those placeholders/variables that,. somehow, seems to suggest the will of abstract the actual "bom" (i.e. different artifactId/groupId), and I'm not sure why

pefernan:
@gitgabrio I was expecting more changes here, shouldn't update more examples?

Eg: https://github.com/apache/incubator-kie-kogito-examples/blob/main/kogito-quarkus-examples/process-usertasks-quarkus/pom.xml#L39

gitgabrio:
Hi @pefernan
I agree, I could have modified examples much more, and we discussed a bit the same topic with @yesamer. 
Point is that the scope of those examples is still fuzzy: are those to be used as starting point for new projects, i.e. as "templates" ? Are those sort of "demo" of all the features we expose ? Depending on that, they could be refactored one way or the other.
There has been already some attempt, in the past, to clarify that, but at the end we did not get to a conclusion.
So, for the moment being, I would prefer to leave them as unchanged as possible, and to iterat over them later on, when/if we could get to some agreement.
Does this make sense ?

pefernan (approved):
great work @gitgabrio! It's clear...  I think we can merge and adapt the examples later when time allows us

kogito-springboot-examples/pom.xml
```diff
@@ -35,77 +35,23 @@
...
   <properties>
+    <!-- Used to define which poms are allowed to have dependencyManagement sections. This is to enforce the convention that only the root pom should have dependencyManagement, and all other poms should inherit from it. -->
+    <allowedPomsList>org.kie.kogito.examples:kogito-springboot-examples</allowedPomsList>
     <java.module.name>org.kie.kogito.examples.springboot</java.module.name>
-    <!-- Override Netty to 4.2.x for these SB examples. The shared kogito-dependencies-bom pins 4.1.x
-         for Quarkus/Vert.x, but Infinispan 15.x's Hot Rod client needs 4.2 (uses io.netty.channel.IoHandle).
-         Per-artifact pins below are required because BOM-imported entries cannot override the parent-
-         imported BOM's explicit versions. Mirrors the matching block in kogito-runtimes' springboot/pom.xml.
-         TODO drop after the planned BOM split. -->
+    <!-- TODO: remove the following properties and declare  kogito-apps-springboot-bom as parent, to have everything already inherited-->
```
martinweiler, line 41: Why is this `TODO` still in here? Isn't `kogito-apps-springboot-bom` imported further down in this file?

martinweiler (approved):
LGTM @gitgabrio - I only had a minor question.

kogito-springboot-examples/pom.xml
yesamer: @gitgabrio ^^

gitgabrio: @martinweiler The idea is to declare `kogito-apps-springboot-bom` as parent, because when boms are "imported" some things get not inherited (properties, build configuration, etc)

PR #2206, opened 2026-04-20
[incubator-kie-issues#2164] Implement quarkus example to verify loading properties from YaML

### LINKED ISSUE #2164: Kogito-quarkus: application.yml files unsupported
A user reported that application.yml file is ignored inside quarkus-based kogito application.

The problem is the same as for https://github.com/apache/incubator-kie-issues/issues/2163, but the executed codepath is different

Scope of this task is to also manage application.yml files inside quarkus-based applications

gitgabrio:
Fixes https://github.com/apache/incubator-kie-issues/issues/2164

This PR clone a process+decisions+rules quarkus example to use application.yaml property file

Depends on:
https://github.com/apache/incubator-kie-drools/pull/6674
https://github.com/apache/incubator-kie-kogito-runtimes/pull/4260


jstastny-cz (approved):
The example including the setup looks good overall.

My only doubt is if the bug really deserves duplication of an example.

gitgabrio:
Hi @jstastny-cz 
As I see them, examples are  place where we demonstrate "features" to overall audience, and that's why I keep creating some of them when something really "new" appear. But this is just my POV

pefernan (approved):

yesamer (approved):

PR #2244, opened 2026-08-28
[incubator-kie-6902] Inherit the platform versions from the kie stack BOMs (Quarkus and Spring Boot examples)

nrknithin:
**Issue:** https://github.com/apache/incubator-kie/issues/6902
**Related:** apache/incubator-kie#6903 (Quarkus 3.33.3.1 upgrade) 

## Why

Every Quarkus example pinned Quarkus 3.27.5.1 in its own pom (version properties, its own `quarkus-bom`
import, explicit `quarkus-maven-plugin` version). Two problems:
- Each Quarkus upgrade in incubator-kie needed a paired version-bump PR here — a cross-repo sync the CI
  no longer supports (each repo's PR builds the other repo's `main`).
- Against the 3.33-built kie jars the pinned 3.27 internals mix with the newer Hibernate / Kafka /
  MongoDB / ProtoStream versions and ten modules fail — the red examples job on #6903.

`kogito-quarkus-bom` already defines `version.io.quarkus` and manages `quarkus-maven-plugin` (bom-splitting
design). The examples only *imported* it, which carries dependency management but not the plugin version
or the property — so they were never able to inherit the version.

## What changed

- `kogito-quarkus-examples/pom.xml`: parent is now `org.kie.kogito:kogito-quarkus-bom`;
  `kogito-springboot-examples/pom.xml`: parent is now `org.kie.kogito:kogito-spring-boot-bom`. Each
  aggregator keeps its apps BOM import and carries the examples' shared build configuration (test
  wiring, container image and resource properties, enforcer skip, compiler release, jar/source
  manifests, reproducible-build settings, project metadata) — a pom has a single parent, so that
  configuration cannot come from the examples root anymore. The two copies are marked to be kept in sync.
- Examples root `pom.xml`: reduced to a plain module list (metadata, reproducible-build settings,
  enforcer skip); nothing else inherits its build configuration any longer (the Java examples are
  standalone poms, the Gradle wrappers only run Gradle).
- 67 Quarkus example poms: pinned properties, per-module `quarkus-bom` import and plugin `<version>`
  deleted; the `quarkus-maven-plugin` groupId is written literally (`io.quarkus`) and the now-unused
  `quarkus.platform.group-id` / `quarkus.platform.artifact-id` properties dropped — with a property as the
  plugin groupId Maven cannot see the managed version during raw-model validation and warns on every
  module. 19 Spring Boot example poms: `spring-boot-maven-plugin` `<version>` deleted. The Spring Boot
  aggregator's own `version.org.springframework.boot` and `version.io.netty` pins are gone too (both now
  inherited from the BOM).
- `onboarding-example/pom.xml`: its version-less `pluginManagement` entry for `quarkus-maven-plugin`
  removed (it shadowed the BOM's entry and made those modules fall back to the latest plugin).

No Quarkus or Spring Boot version is written anywhere in the Maven examples; no incubator-kie change needed.
The only remaining literals are the four Gradle examples' `gradle.properties` (`quarkusPluginVersion` /
`quarkusPlatformVersion` 3.27.5.1, `springBootVersion` 4.0.7): Gradle cannot inherit from a Maven parent,
and the `gradle-examples` CI job does not drive Gradle, so they are inert in CI. They still need a manual
bump once #6903 merges (2 lines × 2 files for Quarkus).

## Effect

- On today's `main`: resolves to the same 3.27.5.1 — no functional change.
- After #6903: the examples pick up 3.33.3.1 automatically; future Quarkus upgrades can't break the
  examples through version mismatches anymore.

## Verification

- Effective poms before/after for a Quarkus, a Spring Boot, a Java and a Gradle example: build/test
  configuration identical for Quarkus and Spring Boot (only additions: the inherited BOM properties and,
  for Spring Boot, the BOM's four compile-scope dependencies that were already on the classpath — resolved
  runtime dependencies identical, 245/245). The Java examples are standalone poms and the Gradle wrappers
  never used the removed root configuration (effective poms unchanged / only kie defaults visible).
- Root reactor, full build with unit + integration tests against the 3.33-built kie jars (the post-#6903
  state): **81/81 modules, 152 suites, 0 failures**; all 62 `quarkus:build` executions on 3.33.3.1 from
  the BOM, including the ten modules that fail on #6903's examples job today; zero
  `plugin.version is missing` warnings.
- Spring Boot examples (`-Dspringboot`, with unit + integration tests — note the `kogito-springboot-examples`
  CI job only builds the aggregator pom, as that profile is not active by default): **35/35 modules,
  0 failures**, every `spring-boot` plugin run on 4.0.7 from the BOM.

### DIFF
--- a/kogito-quarkus-examples/kogito-travel-agency/extended/pom.xml
+++ b/kogito-quarkus-examples/kogito-travel-agency/extended/pom.xml
@@ -35,20 +35,11 @@
     <module>visas</module>
   </modules>
   <properties>
-    <quarkus-plugin.version>3.27.5.1</quarkus-plugin.version>
     <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
     <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
-    <quarkus.platform.version>3.27.5.1</quarkus.platform.version>
   </properties>
   <dependencyManagement>
     <dependencies>
-      <dependency>
-        <groupId>${quarkus.platform.group-id}</groupId>
-        <artifactId>${quarkus.platform.artifact-id}</artifactId>
-        <version>${quarkus.platform.version}</version>
-        <type>pom</type>
-        <scope>import</scope>
-      </dependency>
     </dependencies>
   </dependencyManagement>
 </project>
--- a/kogito-quarkus-examples/kogito-travel-agency/extended/pom.xml
+++ b/kogito-quarkus-examples/kogito-travel-agency/extended/pom.xml
@@ -35,8 +35,6 @@
     <module>visas</module>
   </modules>
   <properties>
-    <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
-    <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
   </properties>
   <dependencyManagement>
     <dependencies>
--- a/kogito-quarkus-examples/kogito-travel-agency/extended/pom.xml
+++ b/kogito-quarkus-examples/kogito-travel-agency/extended/pom.xml
@@ -34,10 +34,4 @@
     <module>travels</module>
     <module>visas</module>
   </modules>
-  <properties>
-  </properties>
-  <dependencyManagement>
-    <dependencies>
-    </dependencies>
-  </dependencyManagement>
 </project>
--- a/kogito-quarkus-examples/kogito-travel-agency/pom.xml
+++ b/kogito-quarkus-examples/kogito-travel-agency/pom.xml
@@ -35,8 +35,6 @@
     <module>extended</module>
   </modules>
   <properties>
-    <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
-    <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
   </properties>
   <dependencyManagement>
     <dependencies>
--- a/kogito-quarkus-examples/kogito-travel-agency/pom.xml
+++ b/kogito-quarkus-examples/kogito-travel-agency/pom.xml
@@ -34,12 +34,6 @@
     <module>basic</module>
     <module>extended</module>
   </modules>
-  <properties>
-  </properties>
-  <dependencyManagement>
-    <dependencies>
-    </dependencies>
-  </dependencyManagement>
   <dependencies>
     <dependency>
       <groupId>io.quarkus</groupId>
--- a/kogito-quarkus-examples/kogito-travel-agency/pom.xml
+++ b/kogito-quarkus-examples/kogito-travel-agency/pom.xml
@@ -35,20 +35,11 @@
     <module>extended</module>
   </modules>
   <properties>
-    <quarkus-plugin.version>3.27.5.1</quarkus-plugin.version>
     <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
     <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
-    <quarkus.platform.version>3.27.5.1</quarkus.platform.version>
   </properties>
   <dependencyManagement>
     <dependencies>
-      <dependency>
-        <groupId>${quarkus.platform.group-id}</groupId>
-        <artifactId>${quarkus.platform.artifact-id}</artifactId>
-        <version>${quarkus.platform.version}</version>
-        <type>pom</type>
-        <scope>import</scope>
-      </dependency>
     </dependencies>
   </dependencyManagement>
   <dependencies>
--- a/kogito-quarkus-examples/pom.xml
+++ b/kogito-quarkus-examples/pom.xml
@@ -26,7 +26,8 @@
 
   <!-- The Quarkus examples inherit from the Kogito Quarkus BOM so that the Quarkus platform version, the
        quarkus-maven-plugin version and the Quarkus-targeted dependency management are all taken from the
-       single place that defines them on the kie side. No Quarkus version is declared in this repository.
+       single place that defines them on the kie side. No Quarkus version is declared in the Maven examples (the Gradle examples keep their own literal pins in
+       gradle.properties: Gradle cannot inherit from a Maven parent).
 
        A pom has a single parent, so this aggregator does not inherit from the examples root
        (kogito-examples), which is a plain module list. The build configuration below (test wiring,
--- a/kogito-quarkus-examples/pom.xml
+++ b/kogito-quarkus-examples/pom.xml
@@ -338,7 +338,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-resources-plugin</artifactId>
-          <version>2.5</version>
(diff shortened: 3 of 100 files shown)


jomarko (commented):
Looks good to me, I did not manual testing.

kogito-springboot-examples/pom.xml
```diff
@@ -24,26 +24,124 @@
...
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+
+    <version.org.kie.kogito>999-SNAPSHOT</version.org.kie.kogito>
+
+    <surefire.forkCount>1</surefire.forkCount>
+    <failsafe.include>**/*IT.java</failsafe.include>
+    <failsafe.exclude>**/Native*IT.java</failsafe.exclude>
+    <alphanetworkCompilerEnabled>false</alphanetworkCompilerEnabled>
+    <tests.category></tests.category>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
josedee: Duplicate with line 106

kogito-quarkus-examples/kogito-travel-agency/pom.xml
```diff
@@ -35,20 +35,9 @@
...
     <dependencies>
-      <dependency>
-        <groupId>${quarkus.platform.group-id}</groupId>
-        <artifactId>${quarkus.platform.artifact-id}</artifactId>
-        <version>${quarkus.platform.version}</version>
-        <type>pom</type>
-        <scope>import</scope>
-      </dependency>
     </dependencies>
   </dependencyManagement>
```
josedee: This block is empty and could be removed

kogito-quarkus-examples/kogito-travel-agency/pom.xml
```diff
@@ -35,20 +35,9 @@
     <module>extended</module>
   </modules>
   <properties>
-    <quarkus-plugin.version>3.27.5.1</quarkus-plugin.version>
-    <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
-    <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
-    <quarkus.platform.version>3.27.5.1</quarkus.platform.version>
```
josedee: This block is empty and could be removed

kogito-quarkus-examples/kogito-travel-agency/extended/pom.xml
```diff
@@ -35,20 +35,9 @@
...
     <dependencies>
-      <dependency>
-        <groupId>${quarkus.platform.group-id}</groupId>
-        <artifactId>${quarkus.platform.artifact-id}</artifactId>
-        <version>${quarkus.platform.version}</version>
-        <type>pom</type>
-        <scope>import</scope>
-      </dependency>
     </dependencies>
   </dependencyManagement>
```
josedee: Could be removed

kogito-springboot-examples/pom.xml
```diff
@@ -154,9 +269,222 @@
     <pluginManagement>
       <plugins>
         <plugin>
-          <groupId>org.springframework.boot</groupId>
-          <artifactId>spring-boot-maven-plugin</artifactId>
-          <version>${version.org.springframework.boot}</version>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-resources-plugin</artifactId>
+          <version>2.5</version>
```
josedee:
```suggestion
          <version>${version.resources.plugin}</version>
```
I think we need to get it from BOM rather than hardcoding

kogito-springboot-examples/pom.xml
josedee: `<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>`

kogito-quarkus-examples/pom.xml
```diff
@@ -219,27 +337,257 @@
...
+                  <requireJavaVersion>
+                    <version>${maven.compiler.release}</version>
+                  </requireJavaVersion>
+                </rules>
+              </configuration>
+            </execution>
+          </executions>
+          <configuration>
+            <skip>true</skip>
+          </configuration>
```
josedee, line 422: If we are defining the enforcer, why do we need to skip

kogito-springboot-examples/pom.xml
```diff
@@ -154,9 +269,222 @@
...
+                  <requireJavaVersion>
+                    <version>${maven.compiler.release}</version>
+                  </requireJavaVersion>
+                </rules>
+              </configuration>
+            </execution>
+          </executions>
+          <configuration>
+            <skip>true</skip>
+          </configuration>
```
josedee, line 352: Same here

kogito-springboot-examples/pom.xml
nrknithin: Leftover from moving the config out of the root pom. Removed the duplicate, kept the one at line 106.

kogito-quarkus-examples/kogito-travel-agency/pom.xml
nrknithin: Right, those are leftovers from dropping the platform BOM import. Removed the empty blocks — also in onboarding-example/payroll/pom.xml, which had the same leftover but wasn't flagged.

kogito-quarkus-examples/kogito-travel-agency/pom.xml
nrknithin: Same block as above — removed.

kogito-quarkus-examples/kogito-travel-agency/extended/pom.xml
nrknithin: Removed here too, along with the same empty blocks in onboarding-example/payroll/pom.xml.

kogito-springboot-examples/pom.xml
nrknithin: Agreed. Worth noting the 2.5 was already hardcoded in the old examples root pom — this PR only moved the block — but since we now inherit from the kie BOM chain, ${version.resources.plugin} (3.1.0) resolves fine. Applied it here and in kogito-quarkus-examples/pom.xml, which had the same hardcode.

kogito-quarkus-examples/pom.xml
nrknithin: We’re not defining the enforcer here; the skip was already present in the old root POM, and this PR only merged duplicate declarations.

Removing our block doesn’t disable it — it’s inherited from `kie-parent`/`kogito-build-no-bom-parent`, and I verified the rules still run and fail.

I tested skipping only the two incompatible rules; the remaining rules work but expose 4 existing dependency violations that need fixing.

I’ll handle those in a follow-up PR if needed, enable the 3 workable rules, and clean up the now-dead `enforce-versions` config.

kogito-springboot-examples/pom.xml
nrknithin: Same as the Quarkus aggregator — this shows as new only because the shared build config moved out of the root pom; the skip itself is pre-existing.

martinweiler (approved):

tiagobento (approved):

